Issue #3, opened 2020-11-06
Midpoint test fails


baparham:
Running from the current `master` (80ce30b), on Windows, with dart version:
`Dart SDK version: 2.10.0-7.3.beta (beta) (Wed Aug 26 09:46:50 2020 +0200) on "windows_x64"`

Does this test pass for you folks? Perhaps on Linux or on another dart version?

Log output:
```
PS C:\git\turf_dart> dart test
Precompiling executable...
Precompiled test:test.
00:05 +1: test\components\midpoint_test.dart: midpoint
29.704243058393104
251.6519826898071
289.2560026566121, 326.5268244185807, 0
00:06 +9 -1: test\components\midpoint_test.dart: midpoint [E]
  Expected: <true>
    Actual: <false>

  package:test_api                         expect
  test\components\midpoint_test.dart 24:5  main.<fn>

00:06 +17 -1: Some tests failed.
```

Thanks!
I'll keep looking at it to see if I can find a bug in the implementation, or in the test.

lukas-h:
midpoint tests work now, because I added the normalization of latitudes (antimeridian, meridian boundaries [180;-180])

see

`Position.toSigned()`

PR #4, opened 2020-11-07
Port midpoint tests from turf.js

baparham:
Midpoint tests taken from https://github.com/Turfjs/turf commit SHA
23d5cb91d77e0c1e2e903a2252f525797f1d0d09 (the current `master` commit as of writing)

Spawned by test failure referenced in issue  #3 

Basically a direct port of the turf.js tests, with the addition of the existing lat/lng in range check the existing turf_dart test had.

Two of these tests still fail, but perhaps that points to an implementation bug in `midpoint`. It seemed starting with the turf.js tests would help identify if there was in fact a bug in the implementation.

### DIFF
--- a/test/components/midpoint_test.dart
+++ b/test/components/midpoint_test.dart
@@ -1,12 +1,21 @@
 import 'package:test/test.dart';
+import 'package:turf/distance.dart';
 import 'package:turf/helpers.dart';
 import 'package:turf/midpoint.dart';
 
+void checkLatLngInRange(Point result) {
+  _lngRange(num lng) => lng >= -180 && lng <= 180;
+  _latRange(num lat) => lat >= -90 && lat <= 90;
+
+  expect(_lngRange(result.coordinates.lng), true,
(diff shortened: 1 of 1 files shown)


test/components/midpoint_test.dart
```diff
@@ -20,8 +29,139 @@ main() {
...
+    );
+    Point pt2 = Point(
+      coordinates: Position.named(
+        lng: 10,
+        lat: 0,
+      ),
+    );
+    Point result = midpoint(pt1, pt2);
+
+    checkLatLngInRange(result);
```
baparham, line 50: I think this test is actually missing the distance check that the other tests have, which also perhaps could be pulled out into a separate function, like `checkLatLngInRange`. I'll keep the fix on hold in case there is other feedback that should be addressed in another patch.

lukas-h:
Hi @baparham, thanks for the pull request, there is not directly a bug in the implementation but it's due to a lack of normalization of coordinates (meridian/antimeridian boundaries)

Issue #5, opened 2020-11-07
Decide whether to sign degrees-radians conversions


baparham:
I'm porting over some of the `helpers` tests from turf.js and their degree/radians conversions expect that a negative input will yield a negative output, which fails with Dart's modulo operator (which always returns a positive value).

I have a fix to sign the output, but I have a hunch you handled this in another way, perhaps related to `Position.toSigned()` so I'd like some input before proceeding. Signing these function outputs breaks the `destination` test, but leaving them as is breaks newly ported turf.js tests for `destination` I have also been working on.

I'll spend some time trying to convert the ported tests in `destination` to use `Position.toSigned()` to see if that solves the issue as well.

js tests from turf/helpers:
```
...
test('radiansToDegrees', t => {
    t.equal(round(radiansToDegrees(Math.PI / 3), 6), 60, 'radiance conversion PI/3');
    t.equal(radiansToDegrees(3.5 * Math.PI), 270, 'radiance conversion 3.5PI');
    t.equal(radiansToDegrees(-Math.PI), -180, 'radiance conversion -PI');
    t.end();
});

test('radiansToDegrees', t => {
    t.equal(degreesToRadians(60), Math.PI / 3, 'degrees conversion 60');
    t.equal(degreesToRadians(270), 1.5 * Math.PI, 'degrees conversion 270');
    t.equal(degreesToRadians(-180), -Math.PI, 'degrees conversion -180');
    t.end();
});
...
```

Any thoughts or paradigm shifts I need to consider would be appreciated.

lukas-h:
Hi, thanks for your effort!

I see the problem now and it’s all over the place.
I haven’t thought about the modulo operator.

This is also connected https://github.com/Turfjs/turf/issues/1124, because instead of the helper function described in this issue, I created the `Position.toSigned()`.

I somehow dislike `Position.toSigned()`, it’s a really dirty solution, because of the recursion and that it is just available for the basic `CoordinateType`s (`Position`, `BBox`) and not `Point`.

lukas-h:
Could this be a solution for the behavior of the modulo/remainder operator?

https://github.com/dart-lang/sdk/issues/28408#issuecomment-273212205

baparham:
I'd want to do some more tests because I'm nervous about losing precision
with the remainder method. Another solution that gives slightly different
mathematical results (I don't know which we need though) is multiplying the
answer found using the dart modulo operator by the sign of the input.

[...]

PR #6, opened 2020-11-08
Add tests and convert modulo to remainder

baparham:
Adds ported tests for `destination` and for a few of the `helpers` functions.

Also changes places that used the `%` dart modulo operator to use `num.remainder()` instead, which matches the javascript implementation of turf.js.

Still todo (perhaps in later merge request?):

- [ ] Longitude normalization Position.toSigned fails for miniscule precision mismatch
- [ ] point 5000 miles away bearing 90 fails (seems to get a bearing in the wrong direction)
- [ ] Longitude normalization BBox.toSigned fails for miniscule precision mismatch

I'm going to port some more tests to the geoJson library in another merge request I think.


baparham:
related to #5

PR #7, opened 2020-11-10
Create initial dart.yml

baparham:
Initial commit to create Actions to test and ensure dart code is formatted properly


baparham:
OK, it runs `dart format`, `dart analyze` and `dart test` now, even if one of them fails, but the commits should be squashed together when merging.

The `format` and `test` steps fail, but I think that's good for now, until we can either fix an incorrect test or the implementation.

PR #13, opened 2021-01-01
Add geomEach meta function

baparham:
Implements the `geomEach` meta function.

Relates to #12

Issue #12, opened 2020-12-12
Port Meta Tests and Functions


baparham:
# *each
- [x] coordEach (@tobrun)
- [x] featureEach (#24)
- [x] flattenEach (#48 )
- [x] clusterEach (WIP @lukas-h #69)
- [x] segmentEach (WIP @armantorkzaban #70)
- [x] geomEach (#13)
- [x] propEach (#24)

# *reduce
- [x] featureReduce (WIP #49)
- [x] flattenReduce (WIP @armantorkzaban #49)
- [x] clusterReduce (WIP @lukas-h #69)
- [x] coordReduce (WIP @armantorkzaban #49)
- [x] propReduce (WIP #49)
- [x] geomReduce (WIP #49)
- [x] segmentReduce (WIP @armantorkzaban #70)

# other
- [x] coordAll (@armantorkzaban #64)
- [x] getCluster (@lukas-h #69)
- [x] ~~getType (@lukas-h)~~ [--](https://github.com/dartclub/turf_dart/issues/61#issuecomment-1062004895)
- [x] ~~getGeom (@lukas-h)~~ [--](https://github.com/dartclub/turf_dart/issues/61#issuecomment-1062004895)
- [x] getCoord (@armantorkzaban)
- [x] getCoords (@armantorkzaban)

lukas-h:
Hi @baparham, any updates?

baparham:
No 😳 it's been a rather busy year, but I can spend some time on this, I've been wanting to get back into flutter/dart development lately so your ping is timely. I'll block out some time this week to devote here.

lukas-h:
Nice! I'm also slowly starting to work on other parts of the library again, e.g. null-safety support which is more than overdue

tobrun:
The most useful meta function to now work on will be `coordEach`. It's heavily used by other turf functions. Happy to work on that one but might tip my toes in the water with an easy meta function to get started :)

lukas-h:
When implementing the functions, please let's keep in mind that my GeoJSON implementation follows a very strict class representation of the RFC standard. I would prefer to have stricter types for the functions, e.g.:

```patch
- void geomEach(dynamic geoJSON, GeomEachCallback callback) {
+ void geomEach(GeoJSONObject geoJSON, GeomEachCallback callback) {
```
also, see my comments here: #24 

---

here you can see all the classes and their inheritance in a tree diagram:
[...]

PR #24, opened 2021-11-21
[meta] add implementation and test for featureEach and propEach

tobrun:
This PR adds implementation for [propEach](https://github.com/Turfjs/turf/blob/master/packages/turf-meta/index.js#L268-L306) and [featureEach](https://github.com/Turfjs/turf/blob/master/packages/turf-meta/index.js#L363-L398). Setup and docs are conform to #13. For testing, I ported the propEach test from turf.js [here](https://github.com/Turfjs/turf/blob/master/packages/turf-meta/test.js#L125-L133). For featureEach there is no dedicated test. After this PR lands, will look at `coordEach` which unlocks many other non meta functionality.

### DIFF
--- a/lib/src/meta.dart
+++ b/lib/src/meta.dart
@@ -99,3 +99,69 @@ void _forEachGeomInGeometryObject(
     throw Exception('Unknown Geometry Type');
   }
 }
+
+/// Callback for propEach
+typedef PropEachCallback = dynamic Function(
+    Map<String, dynamic>? currentProperties, num featureIndex);
+
+/// Iterate over properties in any [geoJSON] object, calling [callback] on each
+/// iteration. Similar to Array.forEach()
+///
+/// For example:
+///
+/// ```dart
+/// FeatureCollection featureCollection = FeatureCollection(
+///   features: [
+///     point1,
+///     point2,
+///     point3,
+///   ],
+/// );
+/// propEach(featureCollection, (currentProperties, featureIndex) {
+///   someOperationOnEachProperty(currentProperties);
+/// });
+/// ```
+void propEach(GeoJSONObject geoJSON, PropEachCallback callback) {
+  if (geoJSON is FeatureCollection) {
+    for (var i = 0; i < geoJSON.features.length; i++) {
+      if (callback(geoJSON.features[i].properties, i) == false) break;
+    }
+  } else if (geoJSON is Feature) {
+    callback(geoJSON.properties, 0);
+  } else {
+    throw Exception('Unknown Feature/FeatureCollection Type');
+  }
+}
+
+/// Callback for featureEach
+typedef FeatureEachCallback = dynamic Function(
+    Feature currentFeature, num featureIndex);
+
+/// Iterate over features in any [geoJSON] object, calling [callback] on each
+/// iteration. Similar to Array.forEach.
+///
+/// For example:
+///
+/// ```dart
+/// FeatureCollection featureCollection = FeatureCollection(
+///   features: [
+///     point1,
+///     point2,
+///     point3,
+///   ],
+/// );
+/// featureEach(featureCollection, (currentFeature, featureIndex) {
+///   someOperationOnEachFeature(currentFeature);
+/// });
+/// ```
+void featureEach(GeoJSONObject geoJSON, FeatureEachCallback callback) {
+  if (geoJSON is Feature) {
+    callback(geoJSON, 0);
+  } else if (geoJSON is FeatureCollection) {
+    for (var i = 0; i < geoJSON.features.length; i++) {
+      if (callback(geoJSON.features[i], i) == false) break;
+    }
+  } else {
+    throw Exception('Unknown Feature/FeatureCollection Type');
+  }
+}
--- a/test/components/meta_test.dart
+++ b/test/components/meta_test.dart
@@ -41,6 +41,16 @@ Feature<GeometryCollection> geomCollection = Feature<GeometryCollection>(
     ],
   ),
 );
+
+List<GeoJSONObject> collection(Feature feature) {
+  FeatureCollection featureCollection = FeatureCollection(
+    features: [
+      feature,
+    ],
+  );
+  return [feature, featureCollection];
+}
+
 List<GeoJSONObject> featureAndCollection(GeometryObject geometry) {
   Feature feature = Feature(
     geometry: geometry,
@@ -57,6 +67,66 @@ List<GeoJSONObject> featureAndCollection(GeometryObject geometry) {
 }
 
 main() {
+  test('propEach --featureCollection', () {
+    collection(pt).forEach((input) {
+      propEach(input, (prop, i) {
+        expect(prop, {'a': 1});
+        expect(i, 0);
+      });
+    });
+  });
+
+  test('propEach --feature', () {
+    propEach(pt, (prop, i) {
+      expect(prop, {'a': 1});
+      expect(i, 0);
+    });
+  });
+
+  test('propEach --breaking of iterations', () {
+    var count = 0;
+    propEach(multiline, (prop, i) {
+      count += 1;
+      return false;
+    });
+    expect(count, 1);
+  });
+
+  test('featureEach --featureCollection', () {
+    collection(pt).forEach((input) {
+      featureEach(input, (feature, i) {
+        expect(feature.properties, {'a': 1});
+        expect(
+            feature.geometry,
+            Point.fromJson({
+              'coordinates': [0, 0],
+            }));
+        expect(i, 0);
+      });
+    });
+  });
+
+  test('featureEach --feature', () {
+    featureEach(pt, (feature, i) {
+      expect(feature.properties, {'a': 1});
+      expect(
+          feature.geometry,
+          Point.fromJson({
+            'coordinates': [0, 0],
+          }));
+      expect(i, 0);
+    });
+  });
+
+  test('featureEach --breaking of iterations', () {
+    var count = 0;
+    featureEach(multiline, (feature, i) {
+      count += 1;
+      return false;
+    });
+    expect(count, 1);
+  });
+
   test('geomEach -- GeometryCollection', () {
     featureAndCollection(geomCollection.geometry!)
         .forEach((GeoJSONObject input) {


lukas-h (approved):
Looks good!

lukas-h (dismissed):

lib/src/meta.dart
```diff
@@ -126,3 +126,69 @@ void geomEach(dynamic geoJSON, GeomEachCallback callback) {
...
+///     point1,
+///     point2,
+///     point3,
+///   ],
+/// );
+/// propEach(featureCollection, (currentProperties, featureIndex) {
+///   someOperationOnEachProperty(currentProperties);
+/// });
+/// ```
+void propEach(dynamic geoJSON, PropEachCallback callback) {
```
lukas-h, line 151: Couldn't we add a more specific type specification, like `GeoJSONObject`?

---
Because my GeoJSON implementation follows the RFC so strictly, we can use the abstract classes: `GeoJSONObject`, as well as `Geometry` and `GeometryType`.

lib/src/meta.dart
```diff
@@ -126,3 +126,69 @@ void geomEach(dynamic geoJSON, GeomEachCallback callback) {
...
+///     point1,
+///     point2,
+///     point3,
+///   ],
+/// );
+/// featureEach(featureCollection, (currentFeature, featureIndex) {
+///   someOperationOnEachFeature(currentFeature);
+/// });
+/// ```
+void featureEach(dynamic geoJSON, FeatureEachCallback callback) {
```
lukas-h, line 186: same here

lib/src/meta.dart
lukas-h: See #12

baparham: #29 has at least one way to accomplish this, which could be replicated here if we think that is an acceptable way to strongly type these functions

tobrun:
Tried updating this PR with latest changes. @lukas-h @baparham lmk what you think.
Also added a test for featureEach, wanted to at least verify if it worked :)

lib/src/meta.dart
```diff
@@ -99,3 +99,66 @@ void _forEachGeomInGeometryObject(
...
+///     point2,
+///     point3,
+///   ],
+/// );
+/// propEach(featureCollection, (currentProperties, featureIndex) {
+///   someOperationOnEachProperty(currentProperties);
+/// });
+/// ```
+void propEach(GeoJSONObject geoJSON, PropEachCallback callback) {
+  var i;
```
lukas-h, line 125: this seems to me to be redundant?

lib/src/meta.dart
```diff
@@ -99,3 +99,66 @@ void _forEachGeomInGeometryObject(
...
+  if (geoJSON is FeatureCollection) {
+    for (var i = 0; i < geoJSON.features.length; i++) {
+      if (callback(geoJSON.features[i].properties!, i) == false) break;
+    }
+  } else if (geoJSON is Feature) {
+    callback(geoJSON.properties!, 0);
+  }
```
lukas-h, line 132: The already implemented function `geomEach` handles unwanted types with an else block and throws an error, like this:

https://github.com/dartclub/turf_dart/blob/6494415188765c42f434dd01b7fb88c28222cdd0/lib/src/meta.dart#L42

Do we want to have the same behaviour for all meta functions or is there a reason why this is not throwing an error?

CC @baparham

lib/src/meta.dart
```diff
@@ -99,3 +99,66 @@ void _forEachGeomInGeometryObject(
...
+/// });
+/// ```
+void featureEach(GeoJSONObject geoJSON, FeatureEachCallback callback) {
+  if (geoJSON is Feature) {
+    callback(geoJSON, 0);
+  } else if (geoJSON is FeatureCollection) {
+    for (var i = 0; i < geoJSON.features.length; i++) {
+      if (callback(geoJSON.features[i], i) == false) break;
+    }
+  }
```
lukas-h, line 163: same here

lukas-h (commented):
Let's discuss the comments,
otherwise looks good 👍

lib/src/meta.dart
```diff
@@ -99,3 +99,66 @@ void _forEachGeomInGeometryObject(
...
+/// );
+/// propEach(featureCollection, (currentProperties, featureIndex) {
+///   someOperationOnEachProperty(currentProperties);
+/// });
+/// ```
+void propEach(GeoJSONObject geoJSON, PropEachCallback callback) {
+  var i;
+  if (geoJSON is FeatureCollection) {
+    for (var i = 0; i < geoJSON.features.length; i++) {
+      if (callback(geoJSON.features[i].properties!, i) == false) break;
```
baparham, line 128: this more relates to `turf_dart`s handling of null values, but if a geoJSON object doesn't have a props, but get's passed in here, this will just throw a runtime error because there is no meaningful null checks.

In this specific instance, would it better to silently throw away the props object and either break or continue?

The other option (which I am leaning towards as the correct option) is to make the callback function parameter nullable and let the user deal with it, since `A Feature object has a member with the name "properties".  The value of the properties member is an object (any JSON object or a JSON null value).` ([3.2](https://datatracker.ietf.org/doc/html/rfc7946#section-3.2)) and we decided to make it nullable.

lib/src/meta.dart
```diff
@@ -99,3 +99,66 @@ void _forEachGeomInGeometryObject(
     throw Exception('Unknown Geometry Type');
   }
 }
+
+/// Callback for propEach
+typedef PropEachCallback = dynamic Function(
+    Map<String, dynamic> currentProperties, num featureIndex);
```
baparham, line 105: see other comment, I think `currentProperties` should be null. In the .JS implementation this could have been undefined the way that code was written.

test/components/meta_test.dart
```diff
@@ -57,6 +67,29 @@ List<GeoJSONObject> featureAndCollection(GeometryObject geometry) {
...
+        expect(
+            feature.geometry,
+            Point.fromJson({
+              'coordinates': [0, 0],
+            }));
+        expect(i, 0);
+      });
+    });
+  });
+
```
baparham, line 92: so, it looks like it may be difficult to implement the same `meta -- breaking of iterations` test against multiple callback types with dart, but we are missing a test for featureEach similar to https://github.com/Turfjs/turf/blob/master/packages/turf-meta/test.js#L1342

for example line 205 of this file should have added `featureEach` to the list of functions, but I'm not sure how tohandle the strong callback typing in dart. Perhaps that test needs to be refactored into some helper functions so we can properly duplicate that test coverage.

test/components/meta_test.dart
```diff
@@ -57,6 +67,29 @@ List<GeoJSONObject> featureAndCollection(GeometryObject geometry) {
...
 
 main() {
+  test('propEach', () {
+    collection(pt).forEach((input) {
+      propEach(input, (prop, i) {
+        expect(prop, {'a': 1});
+        expect(i, 0);
+      });
+    });
+  });
```
baparham, line 77: I know turf.js doesn't have one, but you could easily add a `Feature` test for `propEach` here too to ensure the implementation for that code path is as expected.

lib/src/meta.dart
baparham: I saw this and thought the same thing, but then checked out turf.js's implementation and they just _do nothing_ so ... ?

I think we can do what we want, and I think it makes sense to throw an error if the geoJSON is not a `Feature` or `FeatureCollection`

I would hope that the `.js` devs coming over to dart will start to think more about silent undefined behaviors and lean more towards explicit error handling, so I suspect that this would line up more with that mentality.

lukas-h: I would implement it with the else-clause with `throw Exception(...)`,
because if you are using a function called `propEach` or `featureEach` it should be clear that you are either dealing with `Feature`s or a `FeatureCollection` and that the other GeoJSON types don't have properties or features in them.

lib/src/meta.dart
lukas-h: I just filed another issue that deals with a similar problem in `geomEach` implementation: https://github.com/dartclub/turf_dart/issues/36

tobrun: > which I am leaning towards as the correct option

I went with the nullable callback

test/components/meta_test.dart
tobrun: Implemented two `--breaking of iterations` tests to cover the use-case. Thanks for flagging!

test/components/meta_test.dart
tobrun: added tests for feature and feature collection independently for both `propEach` as `featureEach`

tobrun:
@lukas-h @baparham appreciate the review! ready for another round

lukas-h (approved):

lib/src/meta.dart
```diff
@@ -99,3 +99,69 @@ void _forEachGeomInGeometryObject(
...
+///   someOperationOnEachProperty(currentProperties);
+/// });
+/// ```
+void propEach(GeoJSONObject geoJSON, PropEachCallback callback) {
+  if (geoJSON is FeatureCollection) {
+    for (var i = 0; i < geoJSON.features.length; i++) {
+      if (callback(geoJSON.features[i].properties, i) == false) break;
+    }
+  } else if (geoJSON is Feature) {
+    callback(geoJSON.properties!, 0);
```
baparham, line 130: I think this `!` bang can be dropped from properties

baparham:
@tobrun just make the `comment coverage on PR` an optional step with `allowed-to-fail: true` for that step I think, then this can get past CI

baparham:
> @tobrun just make the `comment coverage on PR` an optional step with `allowed-to-fail: true` for that step I think, then this can get past CI

OH! never mind, you are running from your fork, so of course the token doesn't have write access to this repo. I'll fix the workflow to allow it to fail, and push it to main so you can get this passed.

lib/src/meta.dart
lukas-h: Oh, I missed that

lukas-h:
Hi @baparham @tobrun what’s the status here?

baparham:
Needs to be rebased so it picks up the new ci flows

baparham:
@tobrun I didn't want to force push into your fork, it seemed inappropriate, so if you can rebase this, that would be great!

tobrun:
> @tobrun I didn't want to force push into your fork, it seemed inappropriate, so if you can rebase this, that would be great!

Thanks for ping, bit focused on the day-to-day job with sickness and OOO time but will jump back into this project asap!

PR #25, opened 2021-11-21
Null-safety implementation

lukas-h:
Hi,
this is my attempt at null-safety.

## GeoJson serialization/deserialization
I tried to follow the official GeoJSON RFC-standard as good as possible, there is still some ambiguity left on how to interpret and  translate it into Dart.
- The attributes of `CoordinateType` and it's child-classes `BBox` and `Position` are **non-nullable**.
  https://github.com/dartclub/turf_dart/blob/315b05f49713598b48781f9390bbed2dfaac7c7f/lib/src/geojson.dart#L35
- [RFC 7946 – 3.1.1](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.1) The optional third attribute (`alt`) for `Position` is also non-nullable and by default == 0. I'm not sure if this is a good way of dealing with it?!
- [RFC 7946 – 3.1.2](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.2) and following: I made the `coordinates` attribute for `Point`, `MultiPoint`, `LineString`, `MultiLineString`, `Polygon` and `MultiPolygon` **nullable**. But I only made the "top-level" type nullable: this means, either you assign `null` to `coordinates` or you have to provide it entirely null-safe.
- **But should it be nullable?** Let's vote on that.
  >"GeoJSON processors MAY interpret Geometry objects with empty \"coordinates\" arrays as null objects." [RFC 7946 – 3.1](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1)

- [RFC 7946 – 3.1.8](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.8) similarly to the other types above, I made the `geometries` attribute for `GeometryCollection` nullable
- [RFC 7946 – 3.2](https://datatracker.ietf.org/doc/html/rfc7946#section-3.2) `Feature`: the `geometry` is also nullable
- [RFC 7946 – 3.3](https://datatracker.ietf.org/doc/html/rfc7946#section-3.3) `FeatureCollection`: the `features` attribute "can be empty" according to the RFC-standard, so I chose to treat it as nullable.

## Ported "Components" from Turf.js

For functions such as `midpoint`, `destination`, etc. we have two implementations: "raw" (which takes just the `Position`) and the other one with `Point`s, which just passes `Point.position` attribute to the "raw" function.
https://github.com/dartclub/turf_dart/blob/315b05f49713598b48781f9390bbed2dfaac7c7f/lib/src/midpoint.dart#L14
https://github.com/dartclub/turf_dart/blob/315b05f49713598b48781f9390bbed2dfaac7c7f/lib/src/midpoint.dart#L6

I don't really think using the bang(!)-operator is a good way of dealing with it, because we give the responsibility to check for `null` again to the user, which is kind of opposed to the whole idea of null-safety.
https://github.com/dartclub/turf_dart/blob/315b05f49713598b48781f9390bbed2dfaac7c7f/lib/src/midpoint.dart#L15

Please @baparham @tobrun, review and correct the migration for the meta functions, ...

---

**Also:** I have still an error with json_serializable, which has some kind of type error, I guess it has something to do with the nested-generic types..

When I try to run: `dart run build_runner build`, this is the output:

```sh
...
[SEVERE] json_serializable:json_serializable on lib/src/geojson.dart:

UnimplementedError: (TypeParameterTypeImpl) T?
[INFO] Running build completed, took 3.8s
[INFO] Caching finalized dependency graph completed, took 36ms
[SEVERE] Failed after 3.8s
```
---

@tobrun @baparham if you could take a look and help me with all of the decisions? 🙂

### DIFF
--- a/lib/src/geojson.dart
+++ b/lib/src/geojson.dart
@@ -258,7 +258,7 @@ abstract class Geometry extends GeoJSONObject {
 }
 
 abstract class GeometryType<T> extends Geometry {
-  T? coordinates;
+  T coordinates;
   GeometryType.withType(this.coordinates, String type) : super.withType(type);
 
   static GeometryType deserialize(Map<String, dynamic> json) {
@@ -283,8 +283,8 @@ abstract class GeometryType<T> extends Geometry {
 
 /// Point, as specified here https://tools.ietf.org/html/rfc7946#section-3.1.2
 @JsonSerializable(explicitToJson: true)
-class Point extends GeometryType<Position?> {
-  Point({this.bbox, Position? coordinates})
+class Point extends GeometryType<Position> {
+  Point({this.bbox, required Position coordinates})
       : super.withType(coordinates, GeoJSONObjectTypes.point);
   factory Point.fromJson(Map<String, dynamic> json) => _$PointFromJson(json);
   @override
@@ -298,14 +298,13 @@ class Point extends GeometryType<Position?> {
   Map<String, dynamic> toJson() => super.serialize(_$PointToJson(this));
 
   @override
-  Point clone() =>
-      Point(coordinates: coordinates?.clone(), bbox: bbox?.clone());
+  Point clone() => Point(coordinates: coordinates.clone(), bbox: bbox?.clone());
 }
 
 /// MultiPoint, as specified here https://tools.ietf.org/html/rfc7946#section-3.1.3
 @JsonSerializable(explicitToJson: true)
-class MultiPoint extends GeometryType<List<Position>?> {
-  MultiPoint({this.bbox, List<Position>? coordinates = const []})
+class MultiPoint extends GeometryType<List<Position>> {
+  MultiPoint({this.bbox, List<Position> coordinates = const []})
       : super.withType(coordinates, GeoJSONObjectTypes.multiPoint);
   factory MultiPoint.fromJson(Map<String, dynamic> json) =>
       _$MultiPointFromJson(json);
@@ -316,15 +315,15 @@ class MultiPoint extends GeometryType<List<Position>?> {
 
   @override
   MultiPoint clone() => MultiPoint(
-        coordinates: coordinates?.map((e) => e.clone()).toList(),
+        coordinates: coordinates.map((e) => e.clone()).toList(),
         bbox: bbox?.clone(),
       );
 }
 
 /// LineString, as specified here https://tools.ietf.org/html/rfc7946#section-3.1.4
 @JsonSerializable(explicitToJson: true)
-class LineString extends GeometryType<List<Position>?> {
-  LineString({this.bbox, List<Position>? coordinates = const []})
+class LineString extends GeometryType<List<Position>> {
+  LineString({this.bbox, List<Position> coordinates = const []})
       : super.withType(coordinates, GeoJSONObjectTypes.lineString);
   factory LineString.fromJson(Map<String, dynamic> json) =>
       _$LineStringFromJson(json);
@@ -335,14 +334,14 @@ class LineString extends GeometryType<List<Position>?> {
 
   @override
   LineString clone() => LineString(
-      coordinates: coordinates?.map((e) => e.clone()).toList(),
+      coordinates: coordinates.map((e) => e.clone()).toList(),
       bbox: bbox?.clone());
 }
 
 /// MultiLineString, as specified here https://tools.ietf.org/html/rfc7946#section-3.1.5
 @JsonSerializable(explicitToJson: true)
-class MultiLineString extends GeometryType<List<List<Position>>?> {
-  MultiLineString({this.bbox, List<List<Position>>? coordinates = const []})
+class MultiLineString extends GeometryType<List<List<Position>>> {
+  MultiLineString({this.bbox, List<List<Position>> coordinates = const []})
       : super.withType(coordinates, GeoJSONObjectTypes.multiLineString);
   factory MultiLineString.fromJson(Map<String, dynamic> json) =>
       _$MultiLineStringFromJson(json);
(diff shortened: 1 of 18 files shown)


tobrun:
> RFC 7946 – 3.1.1 The optional third attribute (alt) for Position is also non-nullable and by default == 0. I'm not sure if this is a good way of dealing with it?!

Capturing from the spec:

```
```

tobrun:
> RFC 7946 – 3.1.1 The optional third attribute (alt) for Position is also non-nullable and by default == 0. I'm not sure if this is a good way of dealing with it?!

Capturing from the spec:

`Altitude or elevation MAY be included as an optional third element.`

To me optional equals nullable. The issue with defaulting to 0 means you can't differentiate when a value is uninitialized or actually 0. 

> RFC 7946 – 3.1.2 and following: I made the coordinates attribute for Point, MultiPoint, LineString, MultiLineString, Polygon and MultiPolygon nullable. But I only made the "top-level" type nullable: this means, either you assign null to coordinates or you have to provide it entirely null-safe.

My personal opinion would be to make these non-nullable. Is their any language in the specification that goes against this? Conceptually a GeoJSON type is determined by the underlying geometry, if their isn't any geometry, is it really an instance of it? 

> RFC 7946 – 3.3 FeatureCollection: the features attribute "can be empty" according to the RFC-standard, so I chose to treat it as nullable.

:+1:

lukas-h:
> Conceptually a GeoJSON type is determined by the underlying geometry, if their isn't any geometry, is it really an instance of it?

[RFC 7946 – 3.1](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1)
```
GeoJSON processors MAY interpret Geometry objects with empty "coordinates" arrays as null objects.
```

And for Features [I wanted to treat Feature, FeatureCollection, GeometryCollection and the geometry types (Point, Polygon, ...) the same]:
```
A Feature object has a member with the name "geometry". 
The value of the geometry member SHALL be either a Geometry object as defined above or,
in the case that the Feature is unlocated, a JSON null value.
```
[RFC 7946 – 3.2](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1)

lukas-h:
> To me optional equals nullable. The issue with defaulting to 0 means you can't differentiate when a value is uninitialized or actually 0.

Totally makes sense, but I have to see how that will complicate the implementation of bounding boxes.

Following the RFC, they can have either 4 (without altitude, 2D) or 6 parameters (with alt, 3D), see [here](https://datatracker.ietf.org/doc/html/rfc7946#section-5).

I had implemented it before as 3D by default.
https://github.com/dartclub/turf_dart/blob/14e9b66876de253a5e530ff5a976d5f8970602df/lib/src/geojson.dart#L211

lib/src/geojson.dart
```diff
@@ -403,8 +408,8 @@ class GeometryCollection extends Geometry {
   GeometryCollection({this.bbox, this.geometries = const []})
       : super.withType(GeoJSONObjectTypes.geometryCollection);
   @override
-  BBox bbox;
-  List<GeometryType> geometries;
+  BBox? bbox;
+  List<GeometryType>? geometries;
```
baparham, line 412: per https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.8

this should probably just be a non-nullable array. If it has 0 items in it, it's still an array with no entries, rather than null.

lib/src/geojson.dart
```diff
@@ -251,7 +258,7 @@ abstract class Geometry extends GeoJSONObject {
 }
 
 abstract class GeometryType<T> extends Geometry {
-  T coordinates;
+  T? coordinates;
```
baparham, line 261: Hm, so the problem here is that the way this is a union type with all the geometry types `T` we can't have an empty array be a valid value here?

It could be that we make a point to have a non-nullable `coordinates` value, but the other types can have empty arrays

lib/src/geojson.dart
```diff
@@ -276,12 +283,12 @@ abstract class GeometryType<T> extends Geometry {
 
 /// Point, as specified here https://tools.ietf.org/html/rfc7946#section-3.1.2
 @JsonSerializable(explicitToJson: true)
-class Point extends GeometryType<Position> {
-  Point({this.bbox, Position coordinates})
+class Point extends GeometryType<Position?> {
+  Point({this.bbox, Position? coordinates})
```
baparham, line 287: e.g. there is no language in the rfc that suggests a point may have a null coordinates value

baparham (commented):
a few thoughts

lib/src/geojson.dart
```diff
@@ -494,19 +500,19 @@ class Feature<T extends Geometry> extends GeoJSONObject {
 class FeatureCollection<T extends Geometry> extends GeoJSONObject {
   FeatureCollection({this.bbox, this.features = const []})
       : super.withType(GeoJSONObjectTypes.featureCollection);
-  List<Feature<T>> features;
+  List<Feature<T?>>? features;
```
baparham, line 503: this can just be an empty list, rather than a list or null, right? the spec still indicates it shall be a list, but can be empty.

baparham: https://datatracker.ietf.org/doc/html/rfc7946#section-3.3
```
The value of "features" is a JSON array.
   Each element of the array is a Feature object as defined above.  It
   is possible for this array to be empty.
```

lib/src/geojson.dart
lukas-h: I think, I have a good way to handle this:
Let's make `coordinates` non-nullable, but with the security that if the value of `coordinates` is `null` when deserializing a GeoJSON file, it will be replaced by an empty array.

lib/src/geojson.dart
lukas-h: look at the other comment

lukas-h:
I think I now have implemented everything necessary for the migration
and I would  really appreciate to get reviews on my latest changes 😄

baparham (approved):
Looks like you've addressed our questions and decisions, nice work!

lib/src/geojson.g.dart
```diff
@@ -6,155 +6,126 @@ part of 'geojson.dart';
 // JsonSerializableGenerator
```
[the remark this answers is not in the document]
lukas-h: It‘s auto-generated by json_serializable.

This package: https://pub.dev/packages/json_serializable
A code generator for JSON (de)serialization.

lukas-h:
Should I go ahead and merge it?

@baparham @tobrun any objections?

baparham:
History is reversible. At least here it is 😂

Issue #27, opened 2021-11-22
featureEach, remove dynamic, use GeoJSONObject instead


tobrun:
Capturing from https://github.com/dartclub/turf_dart/issues/12#issuecomment-974893978:

Update `featureEach` in meta to use GeoJSONObject instead of dynamic:

```
- void geomEach(dynamic geoJSON, GeomEachCallback callback) {
+ void geomEach(GeoJSONObject geoJSON, GeomEachCallback callback) {
```

PR #29, opened 2021-11-22
Strongly type geomEach function to GeoJSONObject

baparham:


tobrun:
Closes #29

tobrun (approved):

baparham:
tests passed locally and in ci, so I'll see if I can at least refactor this to be more maintainable/readable

baparham:
Question for discussion: How critical should we consider the comment about speed an efficiency in the geomEach turf.js implementation? I'm not sure if turf.js has any benchmarks out there for measuring performance that we could run against our dart implementation to ensure we maintain performance.

The only thing I think I'd change is to basically have a few more `if` branches deeper in my refactored implementation, rather than passing around several unnecessary arguments. I'll push a new commit that does this for thoughts and review.

lib/src/meta.dart
```diff
@@ -127,3 +112,63 @@ void geomEach(GeoJSONObject geoJSON, GeomEachCallback callback) {
...
+  bool isFeature = geoJSON.type == GeoJSONObjectTypes.feature;
+  return isFeatureCollection
+      ? (geoJSON as FeatureCollection).features[index].properties
+      : (isFeature ? (geoJSON as Feature).properties : {});
+}
+
+GeometryObject getGeometryObject(GeoJSONObject geoJSON, int index) {
+  bool isFeatureCollection =
+      geoJSON.type == GeoJSONObjectTypes.featureCollection;
+  bool isFeature = geoJSON.type == GeoJSONObjectTypes.feature;
```
baparham, line 162: it's worth pointing out that we now do this equality test for each iteration of the loop (e.g. for each feature in a feature collection) solely for the purpose of simplifying the interface of these functions.

I do not know the actual performance impacts in dart of doing this though on large feature collections, but wanted to bring this up for discussion anyway.

lukas-h:
https://github.com/Turfjs/turf/blob/d72985ce1a577b42340fed5fc70efe8e4bc8b062/packages/turf-meta/index.js#L523 this remark?

Maybe we could create a similar benchmark suite: https://github.com/Turfjs/turf/blob/master/packages/turf-meta/bench.js

I found this package for example:
https://pub.dev/packages/benchmark

lib/src/meta.dart
baparham: it seems to me that polymorphism is the answer hidden in here somewhere that lets us simplify this `geomEach` function while still doing as few single line operations on each iteration as possible to hyper-optimize performance.

lukas-h: we could rewrite it like this, but it still does not feel like a *super huge* improvement. At least we wouldn't need the local variables anymore:
```dart
if(geoJSON is FeatureCollection) {
    return geoJSON.features[index].geometry;
} else if (geoJSON is Feature) {
    return geoJSON.geometry;
} else {
    return geoJSON;
}
```

or in a shorter form:
```dart
return geoJSON is FeatureCollection
      ? geoJSON.features[index].geometry
      : (geoJSON is Feature ? geoJSON.geometry : geoJSON);
```

baparham: I think general what I was thinking might make this "weird logic" taste better is splitting the for loop up into more functions that geomEach calls, e.g. `_forEachGeomInFeatureCollection`, `_forEachGeomInFeature`, and `_forEachGeomInGeometryObject` in which case I wouldn't need to perform the "is this a feature/featurecollection" checks on each feature in large feature datasets, the `_forEachGeomInFeatureCollection` would be written specifically for the `isFeatureCollection` use case, etc etc

baparham:
I rebased this on top of #25 and squashed the commits down to one

lib/src/meta.dart
```diff
@@ -25,104 +31,117 @@ typedef GeomEachCallback = dynamic Function(
...
+        featureProperties, featureBBox, featureId);
+  }
+}
 
-    for (var g = 0; g < stopG; g++) {
-      geometry = isGeometryCollection
-          ? geometryMaybeCollection.geometries[g]
-          : geometryMaybeCollection;
-      if (geometry == null) {
+void runGeomEachCallbacks(
```
baparham, line 83: this can probably be private

lib/src/meta.dart
```diff
@@ -25,104 +31,117 @@ typedef GeomEachCallback = dynamic Function(
...
-    }
-    // Only increase `featureIndex` per each feature
-    featureIndex++;
+      break;
+    default:
+      throw Exception('Unknown Geometry Type');
   }
 }
+
+int _getFeaturesLength(FeatureCollection featureCollection) {
```
baparham, line 129: I think this should be inlined up above instead now that it's just the featureCollection.features.length

lib/src/meta.dart
baparham: fixed

lib/src/meta.dart
baparham: fixed

baparham:
> Closes #29

Did you mean another issue here? I don't see an issue that this closes, but it relates to #27 at least

baparham:
@lukas-h @tobrun does this refactor work for you guys? Anything you'd like to do differently?

lib/src/meta.dart
```diff
@@ -25,104 +31,109 @@ typedef GeomEachCallback = dynamic Function(
...
+}
+
+void _runGeomEachCallbacks(
+    GeometryType<dynamic> currentGeometry,
+    GeomEachCallback callback,
+    int featureIndex,
+    Map<String, dynamic> featureProperties,
+    BBox? featureBBox,
+    dynamic featureId) {
+  switch (currentGeometry.type) {
```
lukas-h, line 90: In my PR #33, I used the `is` operator instead of checking the `.type` attribute in a switch case, which could save us some lines of code. Don't know if type checking is more efficient than this.

like this:
```dart
if(currentGeometry is GeometryType) {
// ...
} else if(currentGeometry is GeometryCollection) {
// ...
} else {
    throw Exception('Unknown Geometry Type');
}
```

lib/src/meta.dart
```diff
@@ -25,104 +31,109 @@ typedef GeomEachCallback = dynamic Function(
...
+  for (int geometryIndex = 0;
+      geometryIndex < geometryCollectionLength;
+      geometryIndex++) {
+    currentGeometry = _getGeometry(geometryObject, geometryIndex);
+    _runGeomEachCallbacks(currentGeometry, callback, featureIndex,
+        featureProperties, featureBBox, featureId);
+  }
+}
+
+void _runGeomEachCallbacks(
```
lukas-h, line 83: tried to merge this function with `_forEachGeomInGeometryObject` in my PR #33, please give me a feedback.

lukas-h:
> @lukas-h @tobrun does this refactor work for you guys? Anything you'd like to do differently?

Looks much cleaner now! I tried to make it even a little smaller here: #33

baparham:
> > @lukas-h @tobrun does this refactor work for you guys? Anything you'd like to do differently?
> 
> Looks much cleaner now! 😄 I tried to make it even a little smaller here: #33

thanks! approved and merged in here

lukas-h (approved):

PR #33, opened 2021-11-23
simplify geomEach logic even a little more

lukas-h:
I just tried some stuff, I'm not sure if it's good.


baparham:
I like this, the recursion makes this slightly more elegant, thanks!

baparham (approved):

Issue #35, opened 2021-11-23
Add code coverage to PR test suite


baparham:
It would be nice to get insight into code coverage in PRs, especially at this early stage working on the lower level fundamentals of the library

lukas-h:
I'll look for some solutions!

baparham:
`dart test --coverage=coverage` writes the JSONs for coverage out do a directory, but doesn't report anything, so another step using something similar to `instanbul` or other coverage report generator would be needed. I'm surprised dart doesn't have an out of the box simple coverage reporter built in to at least just print some of the json values.

baparham:
https://pub.dev/packages/coverage looks promising

baparham:
seems most tools just say use `lcov` to format to html, which should be installed in our CI environment, but not necessarily locally, but that's probably OK. I'll experiment with a PR and see what it does.

baparham:
...or did you want to take this @lukas-h ? I see you've assigned your self so I don't want to step on toes :-)

lukas-h:
You can do it if you want 😉

PR #38, opened 2021-11-23
Add coverage reporting to pull requests tests

baparham:


lukas-h:
Quick note: you first have to `dart pub global activate coverage` before running it 🙂

baparham:
yeah, I'm kind of just experimenting with what is already activated and/or installed in the CI environment. Looks like I need to install lcov too if we want html, but that should be OK. If not, I am testing out a third party action to comment with the details we care about in the PR (https://github.com/marketplace/actions/code-coverage-report)

baparham:
fixes #35

lukas-h (approved):
wow, that's great!

Issue #23, opened 2021-11-21
Implement override operators +, -, * with vector operations for Position


lukas-h:
https://github.com/dartclub/turf_dart/blob/c0c3fe9475755a311564c36891c2a2cdb9be8a0c/lib/src/geojson.dart#L174

Issue #20, opened 2021-11-20
Assemble multilinestring from linestring


tobrun:
Refs: [geojson.dart#5](https://github.com/dartclub/turf_dart/blob/main/lib/src/geojson.dart#L5)

When browsing the source code, I was noticing a couple TODO. To make the work visible but to also inform users about potential missing feature gaps. It's better to also document them into an issue. This library supports an implementation of the [GeoJSON standard](https://tools.ietf.org/html/rfc7946) and to be fully compatible the code should allow create a [MultiLineString](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.5) object from a collection of [LineStrings](https://datatracker.ietf.org/doc/html/rfc7946#appendix-A.2).

Issue #19, opened 2021-11-20
Assemble multipoint from points


tobrun:
Refs: [geojson.dart#4](https://github.com/dartclub/turf_dart/blob/main/lib/src/geojson.dart#L4)

When browsing the source code, I was noticing a couple TODO. To make the work visible but to also inform users about potential missing feature gaps. It's better to also document them into an issue. This library supports an implementation of the [GeoJSON standard](https://tools.ietf.org/html/rfc7946) and to be fully compatible the code should allow create a [MultiPoint](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.3) object from a collection of [points](https://datatracker.ietf.org/doc/html/rfc7946#appendix-A.1).

lukas-h:
@tobrun, thank you for filing the issues!

PR #43, opened 2021-12-06
Assemble different types, refactor GeoJSON serialization, general improvements

lukas-h:
New constructors:
- [x] `MultiPoint.fromPoints` #19  
- [x] `LineString.fromPoints`  
- [x] `MultiLineString.fromLineStrings` #20  
- [x] `Polygon.fromPoints` #21  
- [x] `MultiPolygon.fromPolygons`  

---

- [x] General refactoring of the code
- [x] convert `GeoJSONObjectTypes` into `enum`
- [x] Statically type `round` helper function & fix associated unit test 
- [x] Implement vector addition, subtraction, dot product, vector product #23 

–––

still to do:

- [x] Unit tests for the new constructors
- [x] Unit tests to cover more of `geojson.dart`


lukas-h:
@tobrun @baparham 

I wanted to generally improve lots of things in this library.

Happy to hear some feedback! 😄

lib/src/geojson.dart
```diff
@@ -273,46 +315,52 @@ class BBox extends CoordinateType {
...
 }
 
 abstract class GeometryType<T> extends GeometryObject {
   T coordinates;
-  GeometryType.withType(this.coordinates, String type) : super.withType(type);
+
+  GeometryType.withType(this.coordinates, GeoJSONObjectType type, {BBox? bbox})
+      : super.withType(type, bbox: bbox);
 
   static GeometryType deserialize(Map<String, dynamic> json) {
```
lukas-h, line 334: TODO: unit test

lib/src/geojson.dart
```diff
@@ -1,33 +1,44 @@
+// ignore_for_file: hash_and_equals
```
tobrun, line 1: any context why we need to ignore this, wouldn't it make sense to provide hash/equals for all geojson implementations?

tobrun (commented):
great work! As mentioned in PR description that unit tests are lacking for changes made but I also think we could scale up the amount of tests for GeoJSON. Potentially some that take json file as input to validate correct serialization.

lib/src/geojson.dart
lukas-h: Thanks for the hint!
I have implemented the hashCode for the the GeoJSON object types, where I already overloaded the == operator.
I will work on hashCode and == overloading for all types in a different PR

lukas-h:
We are right now at 100% coverage of `geojson.g.dart` and at 69.1% for `geojson.dart`.

lukas-h:
I was thinking about updating the new constructors to have mandatory parameters + assertions (example below).
The reason why I did not implement this earlier was, because the RFC allows the geometry types to be empty.

Please lmk what you think.

```diff
- Polygon.fromPoints({BBox? bbox, List<List<Point>> points = const []})
- : super.withType(
+ Polygon.fromPoints({BBox? bbox, required List<List<Point>> points})
+ : assert(points.length >= 3), 
+ super.withType(
```

```diff
- LineString.fromPoints({BBox? bbox, List<Point> points = const []})
- : super.withType(
+ LineString.fromPoints({BBox? bbox, required List<Point> points})
+ : assert(points.length >= 2),
+ super.withType(

**Comment by baparham** [comment]
> I was thinking about updating the new constructors to have required parameters + assertions (example below). The reason why I did not implement this earlier was, because the RFC allows the coordinates of geometry types to be empty.
> 
> Please lmk what you think.
I think it makes sense, while also asking the question "is a incorrect geometry (e.g. polygon with 1 point) expected to be a NOOP, or throw an error? I suppose that may be for the client to decide, in which case we would expect them to handle such an error?

File: lib/src/geojson.dart [hunk 3]
```diff
@@ -176,14 +211,38 @@ class Position extends CoordinateType {
   Position.of(List<num> list)
       : assert(list.length >= 2 && list.length <= 3),
         super(list);
+
   factory Position.fromJson(List<num> list) => Position.of(list);
 
-  // TODO implement override operators +, -, * with vector operations
+  Position operator +(Position p) => Position.of([
+        lng + p.lng,
+        lat + p.lat,
```
baparham, line 219: does wraparound get handled in another location?

baparham: probably in the `Position.of` constructor, but I can't quite follow the call chain here in the github UI

test/components/geojson_test.dart
```diff
@@ -32,6 +36,24 @@ main() {
       expect(() => Position.of([1, 2, 3, 4]).toList(),
           throwsA(isA<AssertionError>()));
     });
+
+    test('Position vector addition', () {
+      expect(Position(1, 2, 3) + Position(1, 2, 3), Position(2, 4, 6));
+    });
```
baparham, line 42: should you also test that you can properly add two high value real world lat/lng and have it properly wrap around? (same for subtraction, and perhaps cross product and dot product?)

test/components/geojson_test.dart
```diff
@@ -171,39 +204,268 @@ main() {
...
+    ]);
+    expect(a.coordinates.first.first.first, Position(1, 2, 3));
+    expect(() => MultiPolygon.fromPolygons(polygons: []),
+        throwsA(isA<AssertionError>()));
+  });
+
+// examples
+// copied from RFC 7946 https://datatracker.ietf.org/doc/html/rfc7946
+// copied from Wikipedia https://en.wikipedia.org/wiki/GeoJSON#Geometries
+  group('Example file', () {
```
baparham, line 459: perhaps just double check the licensing for these code examples to ensure they are compliant with this project's outbound license

test/components/geojson_test.dart
```diff
@@ -171,39 +204,268 @@ main() {
...
+    var dir = Directory('./test/examples');
+    for (var file in dir.listSync(recursive: true)) {
+      if (file is File) {
+        test(file.path, () {
+          var source = (file).readAsStringSync();
+          var json = jsonDecode(source);
+          GeoJSONObject.fromJson(json);
```
baparham, line 466: does this just check for exceptions not being thrown when constructing these objects? I do not see any expects or other checks for these files.

test/components/geojson_test.dart
lukas-h: we definitely should

test/components/geojson_test.dart
lukas-h: Wikipedia seems to be Creative Commons Attribution-ShareAlike License

RFC examples:
```
Code Components extracted from this document must include Simplified BSD License ...
```

test/components/geojson_test.dart
lukas-h: yes, you are right, the tests are currently that dumb.

test/components/geojson_test.dart
lukas-h: added license notices

lib/src/geojson.dart
baparham: yeah, I'm not sure if the logic belongs here or not, it sort of depends on the use cases. Is there a use case for getting positions that are technically out of bounds or should the value we return from _any_ function always be normalized to within the ±90deg/±180deg range?

baparham (approved):
so far looking good. nice work.

lib/src/geojson.dart
lukas-h: Let's move this to a separate discussion, because it would – as you pointed out – affect a lot of different parts of the library

test/components/geojson_test.dart
lukas-h: depends on a decision about the coordinate normalization logic

PR #48, opened 2021-12-14
Add flattenEach function and tests

baparham:
A few todos:

- [x] Figure out how to handle geometry collections. turf.js doesn't seem to handle them in `flattenEach`
- [x] Figure out how to properly make a test for the `breaking of iteration` since the callbacks seem impossible to type together cohesively with the way the .js test implements it with function pointers. Perhaps we just need to break it up into separate tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,15 +21,15 @@ Most of the implementation is a direct translation from [turf.js](https://github
 - [ ] area
 - [ ] bbox
 - [ ] bboxPolygon
-- [x] [bearing](https://github.com/dartclub/turf_dart/blob/master/lib/bearing.dart)
+- [x] [bearing](https://github.com/dartclub/turf_dart/blob/main/lib/bearing.dart)
 - [ ] center
 - [ ] centerOfMass
 - [ ] centroid
-- [x] [destination](https://github.com/dartclub/turf_dart/blob/master/lib/destination.dart)
-- [x] [distance](https://github.com/dartclub/turf_dart/blob/master/lib/distance.dart)
+- [x] [destination](https://github.com/dartclub/turf_dart/blob/main/lib/destination.dart)
+- [x] [distance](https://github.com/dartclub/turf_dart/blob/main/lib/distance.dart)
 - [ ] envelope
 - [ ] length
-- [x] [midpoint](https://github.com/dartclub/turf_dart/blob/master/lib/midpoint.dart)
+- [x] [midpoint](https://github.com/dartclub/turf_dart/blob/main/lib/midpoint.dart)
 - [ ] pointOnFeature
 - [ ] polygonTangents
 - [ ] pointToLineDistance
@@ -117,7 +117,7 @@ Most of the implementation is a direct translation from [turf.js](https://github
 - [ ] triangleGrid
 
 ### Classification
-- [x] [nearestPoint](https://github.com/dartclub/turf_dart/blob/master/lib/nearest_point.dart)
+- [x] [nearestPoint](https://github.com/dartclub/turf_dart/blob/main/lib/nearest_point.dart)
 
 ### Aggregation
 - [ ] collect
@@ -128,17 +128,17 @@ Most of the implementation is a direct translation from [turf.js](https://github
 - [ ] coordAll
 - [ ] coordEach
 - [ ] coordReduce
-- [x] featureEach
+- [x] [featureEach](https://github.com/dartclub/turf_dart/blob/main/lib/src/meta.dart#L157)
 - [ ] featureReduce
-- [x] flattenEach
+- [x] [flattenEach](https://github.com/dartclub/turf_dart/blob/main/lib/src/meta.dart#L181)
 - [ ] flattenReduce
 - [ ] getCoord
 - [ ] getCoords
 - [ ] getGeom
 - [ ] getType
-- [x] geomEach
+- [x] [geomEach](https://github.com/dartclub/turf_dart/blob/main/lib/src/meta.dart#L34)
 - [ ] geomReduce
-- [x] propEach
+- [x] [propEach](https://github.com/dartclub/turf_dart/blob/main/lib/src/meta.dart#L124)
 - [ ] propReduce
 - [ ] segmentEach
 - [ ] segmentReduce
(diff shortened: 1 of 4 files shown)


baparham:
According to the RFC, "To maximize interoperability, implementations SHOULD avoid nested GeometryCollections." [src](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.8)

which means, that we MAY encounter a GeometryCollection returned as the geoJson object in the `geomEach` callback that `flattenEach` uses.

Turf.js doesn't support it, but I think we should handle the case. One idea might be to simply recursively call `flattenEach` with the same arguments, passing it the `GeometryCollection` which `geomEach` would iterate through to call the callback for each flattened feature.

Thoughts?

baparham:
I'll write a test and propose a solution in code so you can see it more clearly.

baparham:
> According to the RFC, "To maximize interoperability, implementations SHOULD avoid nested GeometryCollections." [src](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.8)
> 
> which means, that we MAY encounter a GeometryCollection returned as the geoJson object in the `geomEach` callback that `flattenEach` uses.
> 
> Turf.js doesn't support it, but I think we should handle the case. One idea might be to simply recursively call `flattenEach` with the same arguments, passing it the `GeometryCollection` which `geomEach` would iterate through to call the callback for each flattened feature.
> 
> Thoughts?

oh I see, we've actually implemented this to strictly forbid this since a `GeometryCollection`'s geometries are of type `GeometryType`

with that information then `geomEach`'s return type should probably be a `GeometryType` rather than a `GeoJSONObject` and then `flattenEach` doesn't need to handle the case of if it receives a geom collection from geomEach.

lukas-h:
> oh I see, we've actually implemented this to strictly forbid this since a `GeometryCollection`'s geometries are of type `GeometryType`

Yeah, that‘s one of the advantages of my super strict class hierarchy! 😊 



> with that information then `geomEach`'s return type should probably be a `GeometryType`

That’s true. Good catch!

lukas-h:
Should we add an explicit warning in the deserialization functions for nested GeometryCollections?

lukas-h:
I’ll do a full review in the morning

lib/src/meta.dart
```diff
@@ -165,3 +165,86 @@ void featureEach(GeoJSONObject geoJSON, FeatureEachCallback callback) {
...
+///   someOperationOnEachFeature(currentFeature);
+/// });
+/// ```
+void flattenEach(GeoJSONObject geoJSON, FlattenEachCallback callback) {
+  try {
+    geomEach(geoJSON, (currentGeomObject, featureIndex, featureProperties,
+        featureBBox, featureId) {
+      if (currentGeomObject is GeometryCollection) {
+        // turf.js code didn't really handle the GeometryCollection case
+        // TODO figure out how to support this and what the behavior is supposed
```
lukas-h, line 187: I think, if we support it, we should also support Feature and FeatureCollection

README.md
```diff
@@ -122,17 +128,17 @@ Most of the implementation is a direct translation from [turf.js](https://github
 - [ ] coordAll
 - [ ] coordEach
 - [ ] coordReduce
-- [ ] featureEach
+- [x] featureEach
```
lukas-h, line 131: should we link to the source code?

test/components/meta_test.dart
```diff
@@ -236,37 +262,91 @@ main() {
...
+  });
+
+  test('flattenEach -- Mixed FeatureCollection', () {
+    List<Feature> features = [];
+    List<int> featureIndices = [];
+    List<int> multiFeatureIndicies = [];
+    flattenEach(fcMixed, (currentFeature, index, multiIndex) {
+      features.add(currentFeature);
+      featureIndices.add(index);
+      multiFeatureIndicies.add(multiIndex);
```
lukas-h, line 326: we could also check for the type with `expect(foo, isA<Feature<Point>>());`

lukas-h (approved):

baparham:
> Should we add an explicit warning in the deserialization functions for nested GeometryCollections?

Something in this place? What should do? Throw a hard error? where could we print a warning?

```dart
--- a/lib/src/geojson.dart
+++ b/lib/src/geojson.dart
@@ -570,9 +570,13 @@ class GeometryCollection extends GeometryObject {
             : BBox.fromJson(
                 (json['bbox'] as List).map((e) => e as num).toList(),
               ),
-        geometries: (json['geometries'] as List?)
-                ?.map((e) => GeometryType.deserialize(e))
-                .toList() ??
+        geometries: (json['geometries'] as List?)?.map((e) {
+              if (e is GeometryCollection) {
+                // TODO what action should we take here for nested geometry
+                // collections?
+              }
+              return GeometryType.deserialize(e);
+            }).toList() ??
             const [],
       );
```

test/components/meta_test.dart
baparham: did something close, since it wasn't type narrowing until I looked at the `.geometry` property

README.md
baparham: fixed, but with line numbers. I'm not convinced yet if that's a good idea. It works now, but might break later

lukas-h: I agree that it will break in the future. Maybe we can later link to the API reference on pub.dev https://pub.dev/documentation/turf/latest/

lukas-h:
> Something in this place? What should we do? Throw a hard error? where could we print a warning?

I mean, it will already throw an hard error inside of [geometryType.deserialize](https://github.com/dartclub/turf_dart/blob/bcc7cfbdb84cf14e597f82f4468d3d7cebf1c093/lib/src/geojson.dart#L391):

```dart
default:
    throw Exception('${json['type']} is not a valid GeoJSON type')
```
But I thought we could maybe give a little bit more context, like: "This implementation does not support nested `GeometryCollection`s.", similar to the one in the README.
Maybe let's put a `case GeoJSONObjectType.geometryCollection:` before the default case, with a dedicated warning?


> ```dart
>        geometries: (json['geometries'] as List?)?.map((e) {
> +              if (e is GeometryCollection) {
> +                // TODO what action should we take here for nested geometry
> +                // collections?
> +              }
> ```
We have to be careful here: `e` is not yet deserialized, so it's still a `Map<String, dynamic>`.

lukas-h (approved):

baparham:
I'll resolve the conflicts with this branch next week and then we can try and merge it

lukas-h:
I want to merge this soon, because #https://github.com/dartclub/turf_dart/pull/49 depends on it.

@baparham is it ok, if I take over? Thanks for the great work already 😄 

CC @armantorkzaban

baparham:
Go for it

lib/src/meta.dart
```diff
@@ -324,14 +324,9 @@ typedef FlattenEachCallback = dynamic Function(
 /// ```
 void flattenEach(GeoJSONObject geoJSON, FlattenEachCallback callback) {
   try {
-    geomEach(geoJSON, (currentGeomObject, featureIndex, featureProperties,
-        featureBBox, featureId) {
-      if (currentGeomObject is GeometryCollection) {
```
lukas-h: I removed this if clause, because it's already handled by that geomEach now takes a GeometryType?, which cannot be a GeometryCollection.

CC @baparham

PR #49, opened 2021-12-15
Feature reducers

baparham:
- [x] geomReduce (@baparham)
- [x] need to redo the JS comment for geomReduce (@armantorkzaban)
--- 
- [x] propReduce (@armantorkzaban)
- [x] propReduce tests (@armantorkzaban)

---

- [x] featureReduce (@armantorkzaban)
- [x] featureReduce tests (@armantorkzaban)

---

- [x] flattenReduce (@armantorkzaban)
- [x] flattenReduce tests (@armantorkzaban)

---

- [x] coordReduce (@armantorkzaban)
- [x] coordReduce tests (@armantorkzaban)


lukas-h:
#48 is merged, so nothing stands in the way of implementing the functions! 😄 

CC @armantorkzaban

armantorkzaban:
I am working on this one.

lib/src/meta.dart
```diff
@@ -158,8 +168,8 @@ class _ShortCircuit {
   _ShortCircuit();
 }
 
-/// Iterates over each geometry in [geoJSON], calling [callback] on each
-/// iteration. Similar to Iterable.forEach()
+/// Iterate over each geometry in [geoJSON], calling [callback] on each
+/// iteration. Similar to [List.forEach()]
```
lukas-h, line 172: in this case, I think Iterable.forEach() was better

lib/src/meta.dart
```diff
@@ -282,7 +369,7 @@ typedef FeatureEachCallback = dynamic Function(
     Feature currentFeature, num featureIndex);
```
armantorkzaban, line 369: Should it not be int?

lukas-h:
@armantorkzaban 

We have to make a technical decision about the types of the reducer functions. I thought of two ways to handle it.
1. `dynamic` types:
    The previousValue will be initialized if initialValue==null like in the original JS version and also like in Arman’s and Brad’s initial versions. We will drop the generic types to not limit it to geodata-related data types.

3. generic types: Very clean, but we break a little bit with Turf.js-compatibility: We don’t have the if-clause to initialize previousValue if initialValue==null. I’ve implemented that in the latest commit to this branch.

Let’s decide together!

lukas-h:
We now decided for a different approach, where we don't lose any type information and are entirely compatible with TurfJS!
But in this version, we added another type check to the functions:
 
```dart
      if (previousValue == null && featureIndex == 0 && currentGeometry is T) {
        previousValue = currentGeometry?.clone() as T;
      } else {
```

lukas-h:
@armantorkzaban For the unit tests, please look at geomReduce, Brad did a great job there. I added some lines to test the behavior with `dynamic` types

lukas-h (approved):

PR #52, opened 2021-12-19
Meta function extension methods

lukas-h:
as discussed here: #12 

*each
- [ ] coordEach
- [ ] featureEach
- [ ] flattenEach
- [ ] clusterEach
- [ ] segmentEach
- [ ] geomEach
- [ ] propEach

*reduce
- [ ]  featureReduce
- [ ] flattenReduce
- [ ] clusterReduce
- [ ] coordReduce
- [ ] propReduce
- [ ] geomReduce
- [ ] segmentReduce

other
- [ ] coordAll
- [ ] getCluster


lib/meta.dart
```diff
@@ -1,3 +1,3 @@
 library turf_meta;
 
-export 'src/meta.dart';
+export 'src/meta.dart' show geomEach, propEach, featureEach;
```
baparham, line 3: Do we need to export the callback typings?

lib/src/meta_extensions.dart
```diff
@@ -0,0 +1,56 @@
...
+        callback,
+        featureIndex ?? 0,
+      );
+    } on meta.ShortCircuit {
+      return;
+    }
+  }
+}
+
+extension GeomEachFeatureCollection on FeatureCollection {
```
baparham, line 48: I'm still not sure I understand why the specific implementations for features and geometries are necessary... Doesn't the call at the abstract GeoJSONObject cover all these cases within the implementation of geomEach?

lukas-h: I thought a little bit about it. I did it because of additional optional parameters. But I don't think tey are needed really

lukas-h:
this will be closed in favor of a new attempt

PR #53, opened 2022-01-15
Add implementation for getCoord & getCoords // invariant package

lukas-h:
- [x] Dart implementation
- [x] Convert JSDoc comments 
- [x] Unit tests


lukas-h:
@armantorkzaban let‘s update this PR’s tasks to include the rest of turf invariant TurfJS package too. Otherwise we could do it separately.
But either way, let‘s document which functions are yet to implement of invariant, so we don’t lose track

armantorkzaban:
- [ ] containsNumber () {}
- [ ] geojsonType() {}
- [ ] featureOf() {}
- [ ] collectionOf() {}
- [ ] getGeom() {}
- [ ] getType() {}

PR #50, opened 2021-12-18
[meta] add coordeach implementation

tobrun:
This PR ports the JS implementation of [coordeach](https://github.com/Turfjs/turf/blob/master/packages/turf-meta/index.js#L36) and related [tests](https://github.com/Turfjs/turf/blob/master/packages/turf-meta/test.js#L135).

#### Work remaining:
 - [x] port JS implementation
 - [x] port tests 
 - [ ] adopt approach to be aligned with geomEach (eg. ShortCircuit)
 - [x] support breaking of iteration for geometryCollection
 - [x] update README.md support table

cc @baparham @lukas-h

PR #55, opened 2022-02-12
Added bbox, center and polyline decode

Dennis-Mwea:


Dennis-Mwea:
Am assuming you meant the polyline decoding algorithm. It's from [mapbox polyline](https://www.npmjs.com/package/@mapbox/polyline) js package. Everything else is from turf.js package.

lib/src/coord_each.dart
```diff
@@ -0,0 +1,93 @@
+import 'package:turf/helpers.dart';
+
+coordEach(GeoJSONObject? geoJson, Function(Position, int coordIndex, int featureIndex, int multiFeatureINdex, int geometryIndex) callback,
```
lukas-h, line 3: there is already another approach to implement this at #50

.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Default ignored files
```
lukas-h, line 1: Let's exclude the files from .idea/

lib/src/center.dart
```diff
@@ -0,0 +1,5 @@
+import 'package:turf/helpers.dart';
+
+List<double> center(BBox bBox) {
```
lukas-h, line 3: In this library, we are not using List<double> to represent coordinates, but the type `Position`, defined in lib/src/geojson.dart.

If you could refactor this function to return:
```dart
return Position.named(
    lng: (bBox.lng1 + bBox + lng2) / 2,
    lat: (bBox.lat1 + bBox.lat2) / 2,
);
```

also very **_important_**:
GeoJSON coordinates are stored in this order (!): [**lng**, **lat**, alt].

lib/src/bbox.dart
```diff
@@ -0,0 +1,24 @@
+import 'package:turf/coord_each.dart';
+import 'package:turf/helpers.dart';
+
+BBox bbox(GeoJSONObject geoJson) {
```
lukas-h, line 4: could you create unit tests for that?

lib/src/bbox.dart
```diff
@@ -0,0 +1,24 @@
...
+  var result = [double.infinity, double.infinity, double.negativeInfinity, double.negativeInfinity];
+
+  coordEach(geoJson, (Position coord, _, __, ___, ____) {
+    print(coord);
+    if (result[0] > coord.lat) {
+      result[0] = coord.lat.toDouble();
```
lukas-h, line 10: very important:
GeoJSON bounding boxes are stored in this order (!): [lng1, lat1(, alt1), lng2, lat2(, alt2)].

I think you have lat as the first argument, which is incorrect.

lukas-h (changes requested):
Let's discuss my comments and work on the open tasks.
Thank you for your contributions so far :)

lukas-h:
> Am assuming you meant the polyline decoding algorithm. It's from [mapbox polyline](https://www.npmjs.com/package/@mapbox/polyline) js package. Everything else is from turf.js package.

It's licensed under BSD 3-clause. Should be compatible :)

https://github.com/mapbox/polyline/blob/master/LICENSE

@baparham

lib/src/coord_each.dart
lukas-h: Let's see if we can maybe use this implementation instead, because it's already very well tested.

CC @tobrun

lib/src/polyline.dart
```diff
@@ -0,0 +1,36 @@
+import 'dart:math' as math;
+
+List<List<double>> decodePolyline(String polyline, {int precision = 5}) {
```
lukas-h, line 3: Let's also use our `Position` type here instead of `List<double>`

lib/src/coord_each.dart
Dennis-Mwea: No problem. I'll have a look.

lib/src/center.dart
Dennis-Mwea: I'll do the necessary changes.

Dennis-Mwea:
@lukas-h I will work on the necessary changes. Thanks for your review.

lib/src/coord_each.dart
tobrun: thanks for ping @lukas-h and awesome for contributing @Dennis-Mwea . I was planning to get back to #50 today.
The idea is that we align the coordEach implementation with geomEach setup from @baparham (pending doing some benchmarks). The micro optimizations done on turf-js might not be needed with dart (to be validated).

morvagergely:
@Dennis-Mwea is there any updates about this PR?

lukas-h:
An initial `coordEach` implementation is pushed to main, so I think, there are no blocking factors any more. Right?

@Dennis-Mwea

lib/src/polyline.dart
Dennis-Mwea: Wrapping the `Position` in a `LineString`  will limit the `decodePolyline` usages. I think it would be best to have it just decode the Polyline string and let the user decide what to do with the results.

lukas-h:
Hi @morvagergely,
there are still some changes requested, which are not done yet.

lib/src/coord_each.dart
Dennis-Mwea: Changed to the existing implementation and removed the duplicate

lib/src/center.dart
Dennis-Mwea: Changed to return Position as requested.

lib/src/bbox.dart
Dennis-Mwea: Changed to return points in the correct order

lib/src/bbox.dart
Dennis-Mwea: Done

Dennis-Mwea: Done

lukas-h:
@Dennis-Mwea great 😄  I will review them!

lib/bbox.dart
```diff
@@ -0,0 +1,3 @@
+library bbox;
```
lukas-h, line 1: our namespaces start with `turf_`, could you change it to `turf_bbox`?

lib/polyline.dart
```diff
@@ -0,0 +1,3 @@
+library polyline;
```
lukas-h, line 1: same here: `turf_polyline`

lib/src/bbox.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:turf/helpers.dart';
+import 'package:turf/meta.dart';
+
+BBox bbox(GeoJSONObject geoJson, {Map<String, dynamic>? options}) {
```
lukas-h, line 4: I think we could simplify this a little by creating a parameter `{bool recompute = false}`, we could entirely get rid of `options` then.

It would also simplify the following if clause to something like this:
```dart
if (geoJson.bbox != null && !recompute) {
```

lib/src/center.dart
```diff
@@ -0,0 +1,5 @@
+import 'package:turf/helpers.dart';
+
+Position center(BBox bBox) {
```
lukas-h, line 3: this is very different from the turf.js implementation:
https://github.com/Turfjs/turf/blob/master/packages/turf-center/index.ts

They have a `GeoJSONObject` as parameter and run the `bbox(...)` function on it.
I would prefer this function to be compatible with the JS-one.

lib/src/polyline.dart
```diff
@@ -0,0 +1,38 @@
+import 'dart:math' as math;
+
+import 'package:turf/helpers.dart';
+
+List<Position> decodePolyline(String polyline, {int precision = 5}) {
```
lukas-h, line 5: Looks great, but could you port some unit tests as well?

test/components/bbox.dart
```diff
@@ -0,0 +1,35 @@
+import 'package:test/test.dart';
+import 'package:turf/bbox.dart';
+import 'package:turf/helpers.dart';
+
+main() {
+  test('bbox', () {
```
lukas-h, line 6: In the original turf.js package, they cover many more different testing scenarios, e.g. different geometry types, collections, recompute, and a null-geometry in a feature. 

https://github.com/Turfjs/turf/blob/master/packages/turf-bbox/test.js

Wdyt, should we cover some more testing cases?

lib/src/bbox.dart
```diff
@@ -0,0 +1,29 @@
...
+  if (geoJson.bbox != null && options != null && true != options['recompute']) {
+    return geoJson.bbox!;
+  }
+
+  var result = [double.infinity, double.infinity, double.negativeInfinity, double.negativeInfinity];
```
lukas-h, line 9: My idea would be to put the values already here  into a BBox `BBox(lng1: double.infinity, .....)`, so we can use lat/lng instead of indexes inside the coordEach callback, which gives more clarity imo.

lib/src/polyline.dart
lukas-h: https://github.com/mapbox/polyline/blob/master/test/polyline.test.js

lukas-h (changes requested):
Thank you @Dennis-Mwea, the implementation looks really good already.
But there are still a few things:
- [ ] Some minor changes, I requested
- [ ] Unit tests: there are still a lot of testing cases, we should cover imo. If you don't have enough time to port all of them, I'm happy to take off some work from you.
- [ ] Documentation: We should take the descriptions of what the functions do from turf.js as well.

Thank you for the amount of work, you've already put into this project! 👍 😄

lib/bbox.dart
Dennis-Mwea: Done

lib/center.dart
```diff
@@ -0,0 +1,3 @@
+library center;
```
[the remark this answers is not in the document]
Dennis-Mwea: Done

lib/src/helpers.dart
```diff
@@ -148,3 +150,61 @@ num convertArea(num area,
...
+/// @param {Object} [properties={}] an Object of key-value pairs to add as properties
+/// @param {Object} [options={}] Optional Parameters
+/// @param {Array<number>} [options.bbox] Bounding Box Array [west, south, east, north] associated with the Feature
+/// @param {string|number} [options.id] Identifier associated with the Feature
+/// @returns {Feature<Point>} a Point feature
+/// @example
+/// var point = turf.point([-75.343, 39.984]);
+///
+/// //=point
+Feature<Point> point(Position? coordinates, Map<String, dynamic> properties, {Map<String, dynamic>? options}) {
```
lukas-h, line 167: I don't think, we need this helper method, because we can just use the class constructors for the same:
```dart
var feature = Feature(...);
var point = Point(...);
```

lib/src/helpers.dart
```diff
@@ -148,3 +150,61 @@ num convertArea(num area,
...
+/// @example
+/// var geometry = {
+///   "type": "Point",
+///   "coordinates": [110, 50]
+/// };
+///
+/// var feature = turf.feature(geometry);
+///
+/// //=feature
+Feature<G> feature<G extends GeometryObject, P>(G geom, P? properties, {Map<String, dynamic>? options}) {
```
lukas-h, line 193: same here, I don't these helper functions are necessary in Dart, because we have proper OOP with nice constructors

lib/src/center.dart
```diff
@@ -1,5 +1,32 @@
...
+/// //addToMap
+/// var addToMap = [features, center]
+/// center.properties['marker-size'] = 'large';
+/// center.properties['marker-color'] = '#000';
+Feature<Point> center<P>(GeoJSONObject geoJSON, Map<String, dynamic> options) {
+  final ext = bbox(geoJSON);
+  final x = (ext[0]! + ext[2]!) / 2;
+  final y = (ext[1]! + ext[3]!) / 2;
+
+  return point(Position.named(lat: y, lng: x), options['properties'], options: options);
```
lukas-h, line 31: we should use the constructors `Feature()` and `Point()` here.

lib/src/center.dart
```diff
@@ -1,5 +1,32 @@
...
+///   [-97.508269, 35.463245]
+/// ]);
+///
+/// var center = turf.center(features);
+///
+/// //addToMap
+/// var addToMap = [features, center]
+/// center.properties['marker-size'] = 'large';
+/// center.properties['marker-color'] = '#000';
+Feature<Point> center<P>(GeoJSONObject geoJSON, Map<String, dynamic> options) {
```
lukas-h, line 26: is it actually needed to use `Map<...> options` here and then call `options['properties']` to access the properties? Could we not simply just put `Map<String, dynamic> properties` in the signature of the function here?

lib/src/center.dart
```diff
@@ -1,5 +1,32 @@
+import 'package:turf/bbox.dart';
 import 'package:turf/helpers.dart';
 
-Position center(BBox bBox) {
-  return Position.named(lat: (bBox.lat1 + bBox.lat2) / 2, lng: (bBox.lng1 + bBox.lng2) / 2);
+/// Takes a {@link Feature} or {@link FeatureCollection} and returns the absolute center point of all features.
+///
+/// @name center
```
lukas-h, line 6: You have imported the JSDoc comments here, but in Dart, there is a little different syntax, basically you don't have annotations such as `@name ...` or `@param {...}`, you simply describe what the function does in a sentence and wrap `[ ]` those brackets around a variable, function, or type name.

Like this:
```dart
/// Takes a [Feature] or [FeatureCollection] and returns the absolute center point of all features.
///
```

You can find a documentation on documentation 😄 here: https://dart.dev/guides/language/effective-dart/documentation

lib/src/bbox.dart
```diff
@@ -1,8 +1,8 @@
 import 'package:turf/helpers.dart';
 import 'package:turf/meta.dart';
 
-BBox bbox(GeoJSONObject geoJson, {Map<String, dynamic>? options}) {
-  if (geoJson.bbox != null && options != null && true != options['recompute']) {
+BBox bbox(GeoJSONObject geoJson, {bool recompute = false}) {
```
lukas-h, line 4: Could you add a few lines of documentation here too?

lukas-h (changes requested):
Thanks!

Dennis-Mwea:
@lukas-h Check the requested changes. Let me know if there's anything to be changed. Thanks

lib/src/bbox.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:turf/helpers.dart';
+import 'package:turf/meta.dart';
+
+/// Calculates the bounding box for any [geoJson] object, including [FeatureCollection].
+/// Uses geojson.[bbox] if available and options.[recompute] is not set.
```
lukas-h, line 5: I think you can remove the `options.` and `geojson.`, otherwise great

lib/src/polyline.dart
```diff
@@ -0,0 +1,79 @@
...
+      lat += latitudeChange;
+      lng += longitudeChange;
+
+      coordinates.add(Position.named(lng: (lng / factor).toDouble(), lat: (lat / factor).toDouble()));
+    }
+
+    return coordinates;
+  }
+
+  static encode(List<Position> coordinates, {int? precision}) {
```
lukas-h, line 64: maybe you could specify a return type of String here

lib/src/center.dart
```diff
@@ -0,0 +1,16 @@
+import 'package:turf/bbox.dart';
+import 'package:turf/helpers.dart';
+
+/// Takes a [Feature]/[FeatureCollection] and returns the absolute center point of all features.
+Feature<Point> center<P>(GeoJSONObject geoJSON, {Map<String, dynamic>? options}) {
```
lukas-h, line 5: I don't like the `options` approach here either, because we are losing the type information here too.

maybe something like this would be better:
```dart
{dynamic id, BBox? bbox, Map<String, dynamic>? properties}
```
instead of 
```dart
{Map<String, dynamic>? options}
```

lib/src/geojson.dart
```diff
@@ -35,9 +36,7 @@ abstract class GeoJSONObject {
       };
 
   static GeoJSONObject fromJson(Map<String, dynamic> json) {
-    GeoJSONObjectType decoded = json['type'] is GeoJSONObjectType
```
lukas-h, line 38: All of the reformatting could cause the `dart format` tool to fail our CI checks

lib/src/polyline.dart
```diff
@@ -0,0 +1,79 @@
+import 'dart:math' as math;
+
+import 'package:turf/helpers.dart';
+
+class Polyline {
```
lukas-h, line 5: what about a small description here?

lib/src/polyline.dart
```diff
@@ -0,0 +1,79 @@
...
+
+    return output;
+  }
+
+  static _py2Round(value) {
+    // Google's polyline algorithm uses the same rounding strategy as Python 2, which is different from JS for negative values
+    return (value.abs() + 0.5).floor() * (value >= 0 ? 1 : -1);
+  }
+
+  static List<Position> decode(String polyline, {int precision = 5}) {
```
lukas-h, line 29: small doc comment here as well

lib/src/polyline.dart
lukas-h, line 64: doc comment here as well

lib/src/polyline.dart
```diff
@@ -0,0 +1,79 @@
+import 'dart:math' as math;
+
+import 'package:turf/helpers.dart';
+
+class Polyline {
+  static String _encode(current, previous, factor) {
```
lukas-h, line 6: I'm always a fan of declaring types, but for private functions it's not so important. You decide, if you want to give a little more type info

lib/src/center.dart
```diff
@@ -0,0 +1,16 @@
...
+  final x = (ext[0]! + ext[2]!) / 2;
+  final y = (ext[1]! + ext[3]!) / 2;
+
+  return Feature<Point>(
+    id: options?['id'],
+    bbox: options?['bbox'],
+    properties: options?['properties'],
+    geometry: Point(coordinates: Position.named(lat: y, lng: x)),
+  );
+}
```
lukas-h, line 16: what about a few unit tests here aswell?

test/components/bbox.dart
```diff
@@ -0,0 +1,97 @@
...
+        Position.named(lat: 100.2, lng: 0.2),
+      ],
+    ],
+  ]));
+  final fc = FeatureCollection(features: [pt, line, poly, multiLine, multiPoly]);
+
+  test("bbox", () {
+    // FeatureCollection
+    final fcBBox = bbox(fc);
+    expect(fcBBox, equals([-10, 100, 4, 130]), reason: "featureCollection");
```
lukas-h, line 68: I think, you cannot compare it to just an array, because `bbox()` returns a `BBox` object

lukas-h: but I don't know if it actually works, it could because `BBox` extends `CoordinateType`, which itself implements `Iterable`.

lukas-h (changes requested):
**really** great work on the polyline and the testing!

Just minor stuff :)

lukas-h (changes requested):
.

test/components/bbox.dart
lukas-h: 👍

lukas-h:
We will go ahead and continue with your changes in a feature branch in this repo

CC @armantorkzaban

lib/src/bbox.dart
armantorkzaban: Resolved in #96

lib/src/polyline.dart
armantorkzaban: Resolved in https://github.com/dartclub/turf_dart/pull/96

lib/src/center.dart
armantorkzaban: Resolved in https://github.com/dartclub/turf_dart/pull/96

lib/src/geojson.dart
armantorkzaban: Resolved in https://github.com/dartclub/turf_dart/pull/96

lib/src/polyline.dart
armantorkzaban: Resolved in https://github.com/dartclub/turf_dart/pull/96

lib/src/polyline.dart
armantorkzaban: Resolved in https://github.com/dartclub/turf_dart/pull/96

lib/src/polyline.dart
armantorkzaban: Resolved in https://github.com/dartclub/turf_dart/pull/96

lib/src/center.dart
armantorkzaban: https://github.com/Turfjs/turf/blob/master/packages/turf-center/test.js

lukas-h:
I will go ahead and close this, because we picked up development at #96 and are shortly before merging already.

@Dennis-Mwea thanks for your contributions. Your commits and name will still live in the git history!

PR #62, opened 2022-03-08
ports invariants package

armantorkzaban:
this PR implements _invariants.dart_ without the _unnecessary_ functions listed below:
 
-  containsNumber
-  geojsonType
-  featureOf
-  collectionOf
-  getGeom
-  getType

edit: also Gitpod setup


lib/src/invariant.dart
```diff
@@ -43,24 +45,31 @@ Position getCoord(dynamic coord) {
 /// var coords = turf.getCoords(poly);
 /// //= [[[119.32, -8.7], [119.55, -8.69], [119.51, -8.54], [119.32, -8.7]]]
```
lukas-h, line 46: We should update this comment, because we don't have this namespace `turf.`, instead it's just `getCoords(...)`.
Also the result should look like this: `[[Position(119.32, -8.7), Position(...), ...]]`, instead of plain arrays

test/components/invariant_test.dart
```diff
@@ -18,12 +21,23 @@ main() {
...
+          Position.of([119.32, -8.7])
+        ]),
+        equals([
+          Position.of([119.32, -8.7]),
+          Position.of([119.55, -8.69]),
+          Position.of([119.51, -8.54]),
+          Position.of([119.32, -8.7])
+        ]));
+    expect(() => getCoords(feature1), throwsA(isA<Exception>()));
+    expect(() => getCoords(feature1), throwsA(isA<Exception>()));
```
lukas-h, line 38: this is a duplicate line.

test/components/invariant_test.dart
```diff
@@ -18,12 +21,23 @@ main() {
...
     var coords = getCoords(feature2);
```
lukas-h, line 39: can we do this inside of `expect(..., ...)` in line 40?

lukas-h, line 39: we checked for:
- simple types such as `null` to throw
- list of `Position`s to return
- features with
  - simple geometry types like `Point` to throw
  - geometry types with a list of coordinates like `LineString` to return
- What we should do to have covered everything: create a test case, with a LineString or Polygon, **not wrapped** with a `Feature`

lukas-h (changes requested):
Great job os far :)
Maybe just one test case more as described below and 1-2 typos!

lib/src/invariant.dart
```diff
@@ -1,19 +1,17 @@
...
 import 'package:turf/turf.dart';
 
 ///
-/// Unwrap a coordinate from a Point Feature, Geometry or a single coordinate.
+/// Unwraps a coordinate from a Point, Feature<Point>, and a Position.
 ///
 /// @name getCoord
-/// @param {Array<number>|Geometry<Point>|Feature<Point>} coord GeoJSON Point or an Array of numbers
-/// @returns {Array<number>} coordinates
+/// @param {Position |Point |Feature<Point>} coord GeoJSON Point or an Array of numbers
```
lukas-h, line 7: this `@param`, `@name`, `@returns` do not exist in Dart documentation, these  Javascript specific. In Dart you just write continuous text with the types annotated in square brackets.

Here is an example: https://github.com/dartclub/turf_dart/blob/main/lib/src/meta.dart

This is a guide on writing documentation in Dart:
https://dart.dev/guides/language/effective-dart/documentation

lib/src/invariant.dart
```diff
@@ -34,16 +32,16 @@ Position getCoord(dynamic coord) {
...
+/// Unwrap coordinates from a Feature, Geometry Object or a List
 ///
 /// @name getCoords
-/// @param {Array<any>|Geometry|Feature} coords Feature, Geometry Object or an Array
-/// @returns {Array<any>} coordinates
+/// @param {List<dynamic>|Geometry|Feature} coords Feature, Geometry Object or a List
+/// @returns {List<any>} coordinates
 /// @example
-/// var poly = turf.polygon([[[119.32, -8.7], [119.55, -8.69], [119.51, -8.54], [119.32, -8.7]]]);
+/// var poly = Polygon(coordinates: [[Position(119.32, -8.7), Position(119.55, -8.69), Position(119.51, -8.54), Position(119.32, -8.7)]]);
```
lukas-h, line 41: the same applies here: refactor to Dart doc
and put the comment in a block like this: ` ```dart `, like in line 12

lukas-h (changes requested):
Unit tests much better, but the documentation should conform to dartdoc

lib/src/invariant.dart
```diff
@@ -31,18 +26,27 @@ Position getCoord(dynamic coord) {
   throw Exception("coord must be GeoJSON Point or Position");
 }
 
+/// Unwrap coordinates from a [Feature], [GeometryObject] or a List
```
lukas-h, line 29: wrap [ ] around `List` too

lib/src/invariant.dart
```diff
@@ -31,18 +26,27 @@ Position getCoord(dynamic coord) {
...
+/// 
+/// Gets a List<dynamic>, [GeometryObject] or a [Feature] or a List<dynamic> and
```
lukas-h, line 31: same here with [ ]

lukas-h (changes requested):

lib/src/invariant.dart
```diff
@@ -26,9 +26,9 @@ Position getCoord(dynamic coord) {
   throw Exception("coord must be GeoJSON Point or Position");
 }
 
-/// Unwrap coordinates from a [Feature], [GeometryObject] or a List
+/// Unwrap coordinates from a [Feature], [GeometryObject] or a [List]
 /// 
-/// Gets a List<dynamic>, [GeometryObject] or a [Feature] or a List<dynamic> and
+/// Gets a List<dynamic>, [GeometryObject] or a [Feature] or a [List]<dynamic> and
```
lukas-h, line 31: should be `[List<dynamic>]`

lukas-h (changes requested):

lukas-h (approved):
Great!

lukas-h:
Seems like we still have some formatting issues, this is why the unit tests fail

lukas-h (approved):

lukas-h (approved):

PR #64, opened 2022-03-15
Port coordAll function and test, also: refactor coordEach callback 

armantorkzaban:


armantorkzaban:
can we change the signature of CoordEachCallback's   CoordinateType? currentCoord,
 to Position? currentCoord?

lib/src/meta.dart
```diff
@@ -1,6 +1,7 @@
 import 'geojson.dart';
 
 typedef CoordEachCallback = dynamic Function(
+  //todo: can we change this Position
```
lukas-h, line 4: Yes, let's do it!

lib/src/meta.dart
```diff
@@ -309,3 +310,33 @@ void featureEach(GeoJSONObject geoJSON, FeatureEachCallback callback) {
...
+///
+/// ```dart
+/// var featureColl = FeatureCollection(features:
+/// [Feature(geometry: Point(coordinates: Position(13,15)))
+/// ,Feature(geometry: LineString(coordinates: [Position(1, 2),
+/// Position(67, 50)]))]);
+///
+/// var coords = coordAll(features);
+/// //= [Position(13,15), Position(1, 2), Position(67, 50)]
+/// @Lukas: return Position or CoordinateType?
```
lukas-h, line 327: Let's use comments for that, not directly in the source code.

`Position?` is the right type. Please refactor coordAll and coordEach to use that type!

lukas-h (changes requested):

lib/src/meta.dart
```diff
@@ -66,7 +66,7 @@ void coordEach(GeoJSONObject geoJSON, CoordEachCallback callback,
             : 0;
 
         if (geomType == GeoJSONObjectType.point) {
-          if (callback(coords as CoordinateType, coordIndex, featureIndex,
+          if (callback(coords as Position, coordIndex, featureIndex,
```
lukas-h, line 69: is the typecast necessary anymore?

lukas-h (approved):

lib/src/meta.dart
lukas-h: Please resolve conversations, if it’s done!

lukas-h:
@armantorkzaban when is this PR mergeable?

lukas-h:
@armantorkzaban the merge conflict was only, because in the meantime I refactored coordEach in another branch that got merged first.

Issue #65, opened 2022-03-16
Update functions and packages in README according to turfjs.org Website


lukas-h:
Our overview of functions is outdated in different ways and should be updated:

- [ ] Update the implementation status and the links to the source code
- [ ] Since we started the project, Turf.js has made some changes to its packages and added new features and functions. We should incorporate that as well. https://turfjs.org/docs/ This can easily be done by looking at the menu on the left side of the website

armantorkzaban:
in this draft PR [#67](https://github.com/dartclub/turf_dart/pull/67) I am updating the README.

armantorkzaban:
Waiting for the meta package milestone to be met.

PR #67, opened 2022-03-16
Update functions and packages in README.md

armantorkzaban:
https://github.com/dartclub/turf_dart/issues/65


README.md
```diff
@@ -100,6 +102,17 @@ Any new benchmarks must be named `*_benchmark.dart` and reside in the
 - [ ] shortestPath
 - [ ] unkinkPolygon
 
+### HELPER
```
lukas-h, line 105: We have to decide if we need those helpers, because in Dart we have those amazing constructors, that JS/TS doesn’t have!

lukas-h (commented):
just the one thing, otherwise great, thanks :)

README.md
```diff
@@ -142,8 +144,8 @@ Any new benchmarks must be named `*_benchmark.dart` and reside in the
 - [ ] featureReduce
 - [x] [flattenEach](https://github.com/dartclub/turf_dart/blob/main/lib/src/meta.dart#L181)
 - [ ] flattenReduce
-- [ ] getCoord
-- [ ] getCoords
+- [x] [getCoord](https://github.com/dartclub/turf_dart/blob/0c4522c7f5b71b04c06bb4d758957edab6d749fa/lib/src/invariant.dart#L11)
+- [x] [getCoords](https://github.com/dartclub/turf_dart/blob/0c4522c7f5b71b04c06bb4d758957edab6d749fa/lib/src/invariant.dart#L50)
 - [ ] getGeom
```
lukas-h, line 149: we decided in #12 to get rid of getGeom and getType, because they don't make any sense in Dart

README.md
```diff
@@ -142,8 +144,8 @@ Any new benchmarks must be named `*_benchmark.dart` and reside in the
 - [ ] featureReduce
```
lukas-h, line 144: can you add a check mark to all the functions that are implemented?

lukas-h (changes requested):
- [ ] Task 1 from description of #65: "Update the implementation status and the links to the source code"

I would say we could get rid of the line of code to each functions, but at least a link to the source file.
But most importantly: We should mark all the implemented features as done

lukas-h (approved):

PR #69, opened 2022-03-19
Port cluster functions and tests

lukas-h:
see #12 (getCluster, clusterEach, clusterReduce)

- [x] Port functions
- [x] Port tests
- [x] Port benchmark tests
- [x] Update JS Doc to proper Dart formatting

### DIFF
--- a/benchmark/meta_benchmark.dart
+++ b/benchmark/meta_benchmark.dart
@@ -5,10 +5,12 @@ import 'package:turf/meta.dart';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:turf/src/clusters.dart';
+
 void main() {
-  Point pt = Point.fromJson({
-    'coordinates': [0, 0]
-  });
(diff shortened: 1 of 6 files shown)


benchmark/meta_benchmark.dart
```diff
@@ -120,4 +123,26 @@ void main() {
       featureEach(featureCollection, featureEachNoopCB);
     });
   });
+
```
lukas-h, line 126: Why not put the benchmark tests in a new file?

lukas-h (commented):
- [ ] We need to export the cluster library in one of the top level files in `lib/`, the public facing api. I don’t know if we should add it to `lib/meta.dart` or should create a new `lib/cluster.dart`
- [ ] Let‘s move the cluster benchmark tests to a separate file!
- [ ] Oh and resolve the merge conflicts!

Coding was great!

PR #70, opened 2022-03-19
Port meta segment functions & lineSegment

lukas-h:
- [x] Update documentation to match new combine geometries check
- [x] test `segmentEach`
- [x] test `segmentReduce`
- [x] test `lineSegment`

### DIFF
--- a/lib/src/line_segment.dart
+++ b/lib/src/line_segment.dart
@@ -27,63 +27,163 @@ import 'meta.dart';
 
 FeatureCollection<LineString> lineSegment(GeoJSONObject geoJson) {
   List<Feature<LineString>> features = [];
-  if (geoJson is Point || geoJson is MultiPoint) {
-    throw Exception('Unsupported GeoJSONObject type');
-  }
-
-  flattenEach(
-    geoJson,
-    (Feature currentFeature, int featureIndex, int multiFeatureIndex) {
-      if (currentFeature.geometry is! Point) {
-        features.addAll(lineSegmentFeature(currentFeature));
-      }
-    },
-  );
-
+  segmentEach(geoJson, (currentSegment, featureIndex, multiFeatureIndex,
+      geometryIndex, segmentIndex) {
+    features.add(currentSegment);
+  });
   return FeatureCollection(features: features);
 }
 
-/// Creates Segments from LineString coordinates
-List<Feature<LineString>> createSegments(List<Position> coords, properties) {
-  List<Feature<LineString>> segments = [];
+/// SegmentEachCallback
+typedef dynamic SegmentEachCallback(
+    Feature<LineString> currentSegment,
+    int featureIndex,
+    int multiFeatureIndex,
+    int? geometryIndex,
+    int segmentIndex);
 
-  coords.reduce((previousCoords, currentCoord) {
-    Feature<LineString> segment = Feature<LineString>(
-        geometry: LineString(coordinates: [previousCoords, currentCoord]),
-        properties: properties,
-        bbox: BBox.named(
-            lat1: previousCoords.lat,
-            lat2: currentCoord.lat,
-            lng1: previousCoords.lng,
-            lng2: currentCoord.lng));
+/// Iterates over 2-vertex line segment in any GeoJSON object, similar to Array.forEach()
+/// (Multi)Point geometries do not contain segments therefore they are ignored during this operation.
+///
+/// Takes [FeatureCollection],[Feature] or [GeometryObject] geojson any GeoJSON
+/// @param {Function} callback a method that takes (currentSegment, featureIndex, multiFeatureIndex, geometryIndex, segmentIndex)
+/// @returns {void}
+/// For example:
+///
+/// ```dart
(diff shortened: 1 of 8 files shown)


lib/src/line_segment.dart
```diff
@@ -1,3 +1,128 @@
...
+}
+
+ * Create BBox between two coordinates (faster than @turf/bbox)
+ *
+ * @private
+ * @param {Array<number>} coords1 Point coordinate
+ * @param {Array<number>} coords2 Point coordinate
+ * @returns {BBox} [west, south, east, north]
+
+bbox(coords1: number[], coords2: number[]): BBox {
```
lukas-h, line 117: we don't need these helper methods. We already have the great `BBox` constructor from `src/helpers.dart`

lib/src/line_segment.dart
```diff
@@ -26,112 +25,63 @@ import 'meta.dart';
...
-  if (geoJson is! LineString // todo: 2-vertex [LineString]?
-
-      ||
-      geoJson is! MultiLineString ||
-      geoJson is! MultiPolygon) {
-    throw Exception('Type is not supported');
+  if (geoJson is! LineString &&
+      geoJson is! MultiLineString &&
+      geoJson is! MultiPolygon &&
+      geoJson is! Polygon &&
```
lukas-h, line 33: what about GeometryCollection?

lukas-h: can you maybe check for the opposite?

```dart
geoJson is Point || geoJson is Multipoint
```

lib/src/line_segment.dart
```diff
@@ -0,0 +1,87 @@
...
+///         [1, 1],
+///         [0, 1],
+///         [0, 0],
+///       ],
+///     ];
+/// var segments = lineSegment(polygon);
+/// //addToMap
+/// var addToMap = [polygon, segments]
+
+FeatureCollection<LineString> lineSegment(GeometryObject geoJson) {
```
lukas-h, line 28: GeometryObject is the **wrong type** if you want to accept Feature and FeatureCollection!

lukas-h: This is, because Feature and FeatureCollection inherit from GeoJSONObject

lib/src/line_segment.dart
armantorkzaban: You are right. I will include. 
Another thing about the conditions: since the function does not accept any Point/MultiPoint types I think we will need further check like below, wdyt?:
`FeatureCollection<LineString> lineSegment(GeometryObject geoJson) {
  List<Feature<LineString>> features = [];
  if (geoJson is! LineString &&
      geoJson is! MultiLineString &&
      geoJson is! MultiPolygon &&
      geoJson is! Polygon &&
      geoJson is! Feature &&
      geoJson is! FeatureCollection &&
      geoJson is! GeometryCollection) {
    throw Exception('Unsupported GeoJSONObject type');
  } else if (geoJson is GeometryCollection) {
    for (GeometryType type in geoJson.geometries) {
      if (type is Point || type is MultiPoint) {
        throw Exception('Unsupported GeoJSONObject type');
      }
    }
  } else if (geoJson is FeatureCollection) {
    for (Feature feature in (geoJson as FeatureCollection).features) {
      if (feature.geometry is Point || feature.geometry is MultiPoint) {
        throw Exception('Unsupported GeoJSONObject type');
      }
    }
  } else {
    flattenEach(geoJson,
        (Feature currentFeature, int featureIndex, int multiFeatureIndex) {
      lineSegmentFeature(geoJson, features); //
    });
  }
  return FeatureCollection(features: features);
}`

lib/src/line_segment.dart
lukas-h: GeoJSON Polymorphism (2)

lib/src/line_segment.dart
```diff
@@ -0,0 +1,272 @@
...
+  T? previousValue,
+  Feature<LineString> currentSegment,
+  T? initialValue,
+  int featureIndex,
+  int? multiFeatureIndex,
+  int? geometryIndex,
+  int segmentIndex,
+);
+
+/// Reduce 2-vertex line segment in any GeoJSON object, similar to Array.reduce()
```
lukas-h, line 212: Should be [Iterable.reduce]

lukas-h (commented):
- What about the public facing API? 
- [x] We should add `export 'src/line_segment.dart' show segmentEach, segmentReduce;` to `lib/meta.dart` and also 
- [x] create a file `lib/line_segment.dart` to export all of `lib/src/line_segment.dart`
- [ ] We should write benchmark tests
- [ ]  We should get someone else to review it @armantorkzaban. Because I’m still not 100% sure, if we implemented the behavior correctly for lineSegment and segmentEach..

armantorkzaban:
> We should get someone else to review it @armantorkzaban. Because I’m still not 100% sure, if we implemented the behavior correctly for lineSegment and segmentEach. Perhaps @baparham and @tobrun can give us the review.

baparham:
Sure, I'll take a look later tonight

PR #96, opened 2022-06-15
Added bbox, center and polyline decode - (2nd attempt, from a local branch).

armantorkzaban:
Carries on the great work of @Dennis-Mwea (#55).


test/components/center_test.dart
```diff
@@ -0,0 +1,52 @@
+
```
lukas-h, line 1: This is actually the only thing that keeps it from being merged. Please implement the test before you work on any other PR. Thank you @armantorkzaban

lukas-h (changes requested):
Please implement unit tests for center function

test/components/center_test.dart
armantorkzaban: The test is already there, implemented. Needs Bbox though, in order to work.

armantorkzaban: I think at this point we can merge the two branches and deal with them as one. @lukas-h, what do you think?

lukas-h: We could also try to combine the two branches to one

armantorkzaban: What do you mean by 'combining'? Do you mean here on GH or do you mean merging? or both?

lukas-h:
closed in favor of #99

PR #98, opened 2022-06-15
Ports bbox-polygon package

armantorkzaban:


lib/bbox_polygon.dart
```diff
@@ -0,0 +1,3 @@
+library turn_bbox_polygon.dart;
```
lukas-h, line 1: typo: it's `turf` instead of `turn`

lib/src/bbox_polygon.dart
```diff
@@ -0,0 +1,42 @@
+import '../helpers.dart';
```
lukas-h, line 1: please use a package: import

lib/src/bbox_polygon.dart
```diff
@@ -0,0 +1,42 @@
...
+
+/// Takes a bbox and returns an equivalent [Feature<Polygon>].
```
lukas-h, line 3: `Takes a [BBox]` would be correct, because the class is meant here

lib/src/bbox_polygon.dart
```diff
@@ -0,0 +1,42 @@
...
+Feature<Polygon> bboxPolygon(BBox bbox,
+    {Map<String, dynamic> properties = const {}, dynamic id}) {
+  var west = bbox[0]!;
+  var south = bbox[1]!;
+  var east = bbox[2]!;
+  var north = bbox[3]!;
+
+  if (bbox.length == 6) {
+    throw Exception(
+        "@turf/bbox-polygon does not support BBox with 6 positions");
```
lukas-h, line 19: rename this to `bboxPolygon`, because this is the JS-spelling

lib/src/bbox_polygon.dart
```diff
@@ -0,0 +1,42 @@
...
+        [
+          Position.of(lowLeft),
+          Position.of(lowRight),
+          Position.of(topRight),
+          Position.of(topLeft),
+          Position.of(lowLeft)
+        ]
+      ],
+    ),
+  );
```
lukas-h, line 41: let's set the bbox attr here, which should be bbox from the signature of the function, but cloned:
```dart
bbox: bbox.clone(),
```

test/components/bbox_polygon.dart
```diff
@@ -0,0 +1,59 @@
...
+    },
+  );
+
+  test(
+    "bbox-polygon -- valid geojson",
+    () {
+      var poly = bboxPolygon(Bbox(0, 0, 10, 10));
+      var coordinates = poly.geometry!.coordinates;
+
+      expect((coordinates[0].length == 5), equals(true));
```
lukas-h, line 21: you don't need the extra parentheses around the condition on the left side and also not the `equals(...)` surrounding the true on the right side

test/components/bbox_polygon.dart
```diff
@@ -0,0 +1,59 @@
...
+  test(
+    "bbox-polygon -- Error handling",
+    () {
+      expect(() => bboxPolygon(Bbox(-110, 70, 5000, 50, 60, 3000)),
+          throwsA(isA<Exception>()));
+    },
+  );
+
+  test(
+    "bbox-polygon -- Translate BBox (Issue #1179)",
```
lukas-h, line 38: I think, the issue reference is irrelevant for us here. WDYT?

lukas-h (changes requested):
a few small improvements, generally good.
But please remove the `equals(...`, where you can. Because I don't think it's necessarry!

test/components/bbox_polygon.dart
armantorkzaban: It depends on how we port the packages. I would remove it and leave it to the future users to check for the details.

lukas-h:
closed in favor of #99

PR #99, opened 2022-06-19
Merge bbox-polygon and bbox, center, polyline decode

armantorkzaban:
Takes the features from PRs: bbox-polygon from #98 and bbox, center and polyline decode from #96


lib/src/geojson.dart
```diff
@@ -317,12 +321,30 @@ class BBox extends CoordinateType {
...
+    num? lng1,
+    num? lat1,
+    num? lng2,
+    num? lat2,
+  }) =>
+      BBox(
+        lng1 ?? this.lng1,
+        lat1 ?? this.lat1,
+        lng2 ?? this.lng2,
+        lat2 ?? this.lat2,
```
lukas-h, line 345: what about the altitudes alt1, alt2?

lukas-h (approved):

lib/src/geojson.dart
lukas-h: can be fixed later. go ahead and merge!

README.md
```diff
@@ -76,9 +76,9 @@ Any new benchmarks must be named `*_benchmark.dart` and reside in the
 - [ ] along
 - [ ] area
 - [ ] bbox
```
lukas-h, line 78: Ok. just found after merging, that bbox is not mentioned here

armantorkzaban: done!

lib/src/geojson.dart
armantorkzaban: I added those as well. But I assume that those should be done in other places too.

lukas-h: But you did it after merge?

armantorkzaban: Yes, after merging. 
We need a PR

PR #100, opened 2022-06-19
Added missing alt1, alt2 in BBox constructor, BBox put in README

armantorkzaban:


lib/src/geojson.dart
```diff
@@ -317,12 +321,28 @@ class BBox extends CoordinateType {
...
   num? get alt2 => _is3D ? _items[5] : null;
 
+  BBox copyWith({
+    num? lng1,
+    num? lat1,
+    num? lng2,
+    num? lat2,
+    num? alt1,
+    num? alt2,
+  }) =>
```
lukas-h, line 342: Wait a second:
The original /& default constructor of the BBox has the following signature:

```dart
BBox(
    num lng1,
    num lat1,
    num alt1,
    num lng2, [
    num? lat2,
    num? alt2,
  ]) : super([
          lng1,
          lat1,
          alt1,
          lng2,
          if (lat2 != null) lat2,
          if (alt2 != null) alt2,
        ]);
```

lukas-h (changes requested):

lib/src/geojson.dart
lukas-h: We should follow the same sequence, otherwise we'll have problems

armantorkzaban: I changed the order of the optional params(Altitudes) according to standard and moved on to change the tests. Please check the result.

lukas-h (dismissed):

lukas-h:
2 dimensional: `[lng1, lat1, lng2, lat2]`
3 dimensional: `[lng1, lat1, alt1, lng2, lat2, alt2]`

lukas-h (approved):

Issue #101, opened 2022-06-20
Port feature conversion part I


armantorkzaban:
Feature Conversion

- [x] lineToPolygon
- [x] polygonToLine

and their tests

- [x] lineToPolygon_test
- [x] polygonToLine_test

PR #104, opened 2022-06-20
Ports lineToPolygon, polygonToLine, and their tests

armantorkzaban:
Implements parts of #101


lukas-h:
Should I already give a review?

armantorkzaban:
> Should I already give a review?

Would be nice if you could review the lineToPolygon and its test.

lukas-h:
Just fyi, my incomplete implementation attempt for lineToPolygon. It does not contain "closing" of the polygon (first==last coord), etc.

```dart
import 'package:turf/helpers.dart';
import 'package:turf/meta.dart';

Feature lineToPolygon(GeoJSONObject geom) {
  if (geom is Feature) {
    return _toPolygonFeature(geom);
  } else if (geom is FeatureCollection) {
    return _toPolygonFeatureCollection(geom);
  } else {
    return _toPolygonSimple(geom);
  }
}

Feature _toPolygonFeatureCollection(FeatureCollection coll) {
  var multiPolygon = MultiPolygon(
    coordinates: [],
  );
  geomEach(
    coll,
    (
      GeometryType? currentGeometry,
      int? featureIndex,
      Map<String, dynamic>? featureProperties,
      BBox? featureBBox,
      dynamic featureId,
    ) {
      if (currentGeometry is LineString) {
        multiPolygon.coordinates.add([currentGeometry.coordinates]);
      } else if (currentGeometry is MultiLineString) {
        multiPolygon.coordinates.add(currentGeometry.coordinates);
      } else {
        throw Exception('unsupported type');
      }
    },
  );
  return Feature(geometry: multiPolygon);
}

Feature _toPolygonFeature(Feature feature) {
  return _toPolygonSimple(feature.geometry);
}

Feature _toPolygonSimple(
  GeoJSONObject? geom, [
  Map<String, dynamic>? properties,
  dynamic id,
]) {
  if (geom is LineString) {
    return Feature(
      id: id,
      properties: properties,
      geometry: Polygon(coordinates: [geom.coordinates]),
    );
  } else if (geom is MultiLineString) {
    return Feature(
      id: id,
      properties: properties,
      geometry: Polygon(coordinates: geom.coordinates),
    );
  } else {
    throw Exception('unsupported type');
  }
}
```

lib/src/line_to_polygon.dart
```diff
@@ -0,0 +1,181 @@
...
+Feature<Polygon> lineStringToPolygon(
+    GeoJSONObject line, bool autoComplete, bool orderCoords,
+    {Map<String, dynamic>? properties}) {
+  properties = properties ?? (line is Feature ? line.properties ?? {} : {});
+  var geom = line is LineString ? line : (line as Feature).geometry;
+  List<dynamic> coords = (geom is LineString || geom is MultiLineString)
+      ? (geom is LineString)
+          ? geom.coordinates
+          : (geom as MultiLineString).coordinates
+      : ((geom as Feature).geometry as GeometryType).coordinates;
```
lukas-h, line 119: Please clean that up. The expression is super big and painful to read..

it should be sufficient to have a single if statement:
```dart
if(geom is Feature && geom.geometry is GeometryType) {
    geom = geom.geometry;
}
```

The rest should be dealt with in the if-else if-else clauses below

lib/src/line_to_polygon.dart
```diff
@@ -0,0 +1,181 @@
...
+  var first = coords[0];
+  var x1 = first[0];
+  var y1 = first[1];
+  var last = coords[coords.length - 1];
+  var x2 = last[0];
+  var y2 = last[1];
+  if (x1 != x2 || y1 != y2) {
+    coords.add(first);
+  }
+  return coords;
```
lukas-h, line 171: It is very important to **always** clone a class, otherwise it is a **reference**, not a newly instantiated object...

you can simplify to a great extent here:
```dart
    var newCoords = coords.map((c) => c.clone()).toList();
    if(newCoords.first != newCoords.last){
        newCoords.add(first.clone());
    }
    return newCoords;
```

lib/src/polygon_to_line.dart
```diff
@@ -0,0 +1,69 @@
...
+///   Position.of([145, -20]),
+///   Position.of([125, -20]),
+///   Position.of([125, -30])
+///   ]
+/// ]);
+/// var line = polygonToLine(poly);
+/// //addToMap
+/// var addToMap = [line];
+/// ```
+polygonToLine(GeoJSONObject poly, {Map<String, dynamic>? properties}) {
```
lukas-h, line 22: please specify return type

lib/src/polygon_to_line.dart
```diff
@@ -0,0 +1,69 @@
...
+  var geom = poly is Feature ? poly.geometry : poly;
+
+  (poly is Feature)
+      ? properties = poly.properties
+      : properties = <String, dynamic>{};
```
lukas-h, line 27: either use the `? :` expression properly like this:
```dart
properties = poly is Feature ? poly.properties : {};
```
or use an if-else statement

lib/src/polygon_to_line.dart
```diff
@@ -0,0 +1,69 @@
...
+  if (geom is Polygon) {
+    return _polygonToLine(poly, properties: properties);
+  } else if (geom is MultiPolygon) {
+    return _multiPolygonToLine(poly, properties: properties);
+  } else {
+    throw Exception("invalid poly");
+  }
+}
+
+Feature _polygonToLine(GeoJSONObject poly, {Map<String, dynamic>? properties}) {
```
lukas-h, line 38: why are you not giving this the parameter type `Polygon` and pass `geom` inside `polygonToLine` directly to it?

lib/src/polygon_to_line.dart
```diff
@@ -0,0 +1,69 @@
...
+  var geom = poly is Feature ? poly.geometry : poly;
+  var coords = (geom as GeometryType).coordinates;
+  properties =
+      properties ?? (poly is Feature ? poly.properties : <String, dynamic>{});
+
+  return _coordsToLine(coords, properties);
+}
+
+FeatureCollection _multiPolygonToLine(GeoJSONObject multiPoly,
```
lukas-h, line 47: here please type MultiPolygon instead of `GeoJSONObject`

lukas-h (changes requested):

lukas-h:
BTW, once review requested, please change the PR from draft to a normal PR. @armantorkzaban

lukas-h:
Let's narrow this PR down to just `lineToPolygon` and `polygonToLine`. This way we can merge faster, without having to implement  the rest of the conversion methods here

lukas-h:
What is the status on this one? Looks to me also like an easy to finish thing.
Let's finish this asap, so we can reduce the number of tasks that we work on at the same time.

lukas-h (changes requested):

lukas-h (approved):

Issue #120, opened 2022-07-07
[truncate] Why is element unused in forEach loop?


lukas-h:
Something I missed during code review:

https://github.com/dartclub/turf_dart/blob/2ffce92ba902a2f024b31ce11cc9e0e941945dd6/lib/src/truncate.dart#L73

armantorkzaban:
Dealt with in https://github.com/dartclub/turf_dart/pull/121

PR #121, opened 2022-07-07
Debugs GeometryCollection condition in truncate

armantorkzaban:
#120 The reason the condition for GeometryCollection was not called and the test had not caught it was that we were working with geomEach, which basically goes inside the geometries of a GeomCollection if that is the argument or it will work with the GeomCollection geometry of a Feature, which we didn't have in our test cases.


lib/src/truncate.dart
```diff
@@ -0,0 +1,104 @@
...
+                _truncateCoords(currentCoord!, precision, coordinates);
+          } else if (currentGeometry is MultiLineString) {
+            currentGeometry.coordinates[multiFeatureIndex!][coordIndex!] =
+                _truncateCoords(currentCoord!, precision, coordinates);
+          } else if (currentGeometry is MultiPolygon) {
+            currentGeometry.coordinates[multiFeatureIndex!][geometryIndex!]
+                    [coordIndex!] =
+                _truncateCoords(currentCoord!, precision, coordinates);
+          } else {
+            for (var geojson
```
lukas-h, line 72: we have a naming conflict here.
Already the function signature uses the name `geojson` in line 36.
```dart
void _replaceCoords(int precision, int coordinates, GeoJSONObject geojson) {
```

PR #123, opened 2022-07-23
Added support for the area function

yardenfi:


yardenfi:
@lukas-h Can you take a look please?

lib/src/area.dart
```diff
@@ -0,0 +1,130 @@
...
+    case GeoJSONObjectType.multiPoint:
+    case GeoJSONObjectType.lineString:
+    case GeoJSONObjectType.multiLineString:
+      return 0;
+    case GeoJSONObjectType.geometryCollection:
+      final geometryCollection = geom as GeometryCollection;
+      for (i = 0; i < geometryCollection.geometries.length; i++) {
+        total += calculateArea(geometryCollection.geometries[i]);
+      }
+      return total;
```
lukas-h, line 55: `GeometryCollection`, `Feature`, or `FeatureCollection` can not be the case, because it is based on the `geomEach`, which already loops over all of the concrete `GeometryTypes`, so you don't need to do it here! 
 https://github.com/dartclub/turf_dart/blob/78f68e98c2f392cc3db71311a4825fb60c553085/lib/src/meta/geom.dart#L144

README.md
```diff
@@ -74,7 +74,7 @@ Any new benchmarks must be named `*_benchmark.dart` and reside in the
 
 ### Measurement
 - [ ] along
-- [ ] area
+- [x] area
```
lukas-h, line 77: we want to reference it here with a link to the implementation file, like this:

```md
- [x] [area](https://github.com/dartclub/turf_dart/blob/main/lib/src/area.dart)
```

test/components/area_test.dart
```diff
@@ -0,0 +1,66 @@
+import 'package:test/test.dart';
+import 'package:turf/src/geojson.dart';
+import 'package:turf/turf.dart';
+
+main() {
```
lukas-h, line 5: Please wrap all the tests with a `group('', (){});`

lib/src/area.dart
```diff
@@ -0,0 +1,130 @@
...
+/// JPL Publication 07-03, Jet Propulsion
+/// Laboratory, Pasadena, CA, June 2007 https://trs.jpl.nasa.gov/handle/2014/40409
+num ringArea(List<Position> coords) {
+  var p1;
+  var p2;
+  var p3;
+  var lowerIndex;
+  var middleIndex;
+  var upperIndex;
+  var i;
```
lukas-h, line 95: I'd prefer static types, e.g. `late int foo;`

lib/src/area.dart
```diff
@@ -0,0 +1,130 @@
...
+      p3 = coords[upperIndex];
+      total += (rad(p3[0]) - rad(p1[0])) * sin(rad(p2[1]));
+    }
+
+    total = (total * earthRadius * earthRadius) / 2;
+  }
+  return total;
+}
+
+num rad(num number) {
```
lukas-h, line 128: Please make this radians function private, because I assume, it's not an export of the lib, right?

lib/src/area.dart
```diff
@@ -0,0 +1,130 @@
...
+    total += ringArea(coords[0]).abs();
+    for (var i = 1; i < coords.length; i++) {
+      total -= ringArea(coords[i]).abs();
+    }
+  }
+  return total;
+}
+
+///
+/// Calculate the approximate area of the polygon were it projected onto the earth in square meters.
```
lukas-h, line 80: `[Polygon]` instead of `polygon`. That way the type is highlighted with Dart docs

lib/src/area.dart
lukas-h: and then let's work with a default case that thows an `Exception`

lib/src/area.dart
```diff
@@ -0,0 +1,130 @@
...
+num? area(GeoJSONObject geojson) {
+  return geomReduce<num>(geojson, (value, geom, _, __, ___, ____) {
+    return value! + calculateArea(geom!);
+  }, 0);
+}
+
+/// Calculate Area
+num calculateArea(GeometryType geom) {
+  num total = 0;
+  int i;
```
lukas-h, line 35: Please remove the `i` here.

lib/src/area.dart
```diff
@@ -0,0 +1,130 @@
...
+  switch (geom.type) {
+    case GeoJSONObjectType.polygon:
+      return polygonArea((geom as Polygon).coordinates);
+    case GeoJSONObjectType.multiPolygon:
+      geom as MultiPolygon;
+      for (i = 0; i < geom.coordinates.length; i++) {
```
lukas-h, line 41: And let's instead make this `var i = 0; ...`

lib/src/area.dart
```diff
@@ -0,0 +1,130 @@
...
+      }
+      return total;
+    case GeoJSONObjectType.point:
+    case GeoJSONObjectType.multiPoint:
+    case GeoJSONObjectType.lineString:
+    case GeoJSONObjectType.multiLineString:
+      return 0;
+    case GeoJSONObjectType.geometryCollection:
+      final geometryCollection = geom as GeometryCollection;
+      for (i = 0; i < geometryCollection.geometries.length; i++) {
```
lukas-h, line 52: here as well

lib/src/area.dart
```diff
@@ -0,0 +1,130 @@
...
+    case GeoJSONObjectType.featureCollection:
+      final featureCollection = geom as FeatureCollection;
+      for (i = 0; i < featureCollection.features.length; i++) {
+        total += calculateArea(featureCollection.features[i].geometry as GeometryType);
+      }
+      return total;
+  }
+}
+
+num polygonArea(List<List<Position>> coords) {
```
lukas-h, line 68: Is this a function, that should be used independently or is this solely an implementation function? If we don't want to export it, we  need to make it private with `_`.

And if it shouldn't be private, then I would suggest using `Polygon` instead of `List<List<Position>>`` as the input type

lib/src/area.dart
```diff
@@ -0,0 +1,130 @@
+import 'dart:math';
+
+import '../helpers.dart';
+import '../meta.dart';
+
+/// Takes one or more features and returns their area in square meters.
```
lukas-h, line 6: I would replace `Takes one or more features` with `Takes a [GeoJSONObject]`, because that way we have made use of the type annotation for Dart docs.

lukas-h (changes requested):
Thank you for your great contribution! 😄 

It already looks very good, just a few small changes requested.

lib/src/area.dart
yardenfi: @lukas-h Ok, how would the exception look like?

lukas-h (dismissed):
lgtm!

We can go ahead and squash-merge!

lib/src/area.dart
lukas-h: `throw Exception('unsupported type ${geom.type');`

lukas-h:
Please also run `dart format .` to format the source code properly, so our CI pipeline does not fail the build.
Thank you!

lib/src/area.dart
yardenfi: @lukas-h Is this ok now?

lukas-h (approved):

lib/src/area.dart
lukas-h: yes, lgtm! 😄

lukas-h:
Thank you very much for the contribution 👍 😄

Issue #126, opened 2022-08-10
Implement polygon-smooth algo & tests


lukas-h:

lukas-h:
We'll start with this at the end of the week

armantorkzaban:
I will close this as it is merged.

PR #127, opened 2022-08-21
Implement polygon-smooth and its tests

lukas-h:
#126


armantorkzaban (approved):
Well done. Thank you!
Can't we test with lineIntersect prior to smoothing to warn/prevent creation of degenerate Polygons?

PR #155, opened 2024-01-20
Document properties set by NearestPointOnLine on returned Point

Leffe108:
This is documented on turf.js docs, but was omitted in turf_dart.

These returned properties are quite handy in some situations and saves you doing a bit extra work to find these data.


jsiedentop (approved):

jsiedentop:
Thank you for the contribution. In my opinion, this provides clarity, and if they do the same for turf.js, there's nothing against it.

Issue #157, opened 2024-01-23
lineSlice


leiflinse-trivector:
Just a heads up that I work on `lineSlice` and there will come a PR on that one.

jsiedentop:
Done with https://github.com/dartclub/turf_dart/pull/158

PR #158, opened 2024-01-25
Adds lineSlice method

leiflinse-trivector:
Issue: https://github.com/dartclub/turf_dart/issues/157

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ Any new benchmarks must be named `*_benchmark.dart` and reside in the
 - [ ] [lineIntersect](https://github.com/dartclub/turf_dart/blob/main/lib/src/line_intersect.dart)
 - [ ] lineOverlap
 - [x] [lineSegment](https://github.com/dartclub/turf_dart/blob/main/lib/src/line_segment.dart)
-- [ ] lineSlice
+- [x] lineSlice
 - [ ] lineSliceAlong
 - [ ] lineSplit
 - [ ] mask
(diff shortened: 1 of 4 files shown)


leiflinse-trivector:
Rebased upon main

leiflinse-trivector:
* Rebased upon main (along returns full Feature now)
* Changed parameters to lineSlice to be Feature<point> instead of just Point and updated testes for that
* I fixed the port where I inserted a guard for too short lines that shouldn't be there and caused the interpolation to fail in some cases.

I didn't have the time to retrofit changes due to main rebase back into the history so squashed it all together.

jsiedentop:
Are you finished so that I can do a review?

leiflinse-trivector:
I think it is ready for review.

I have a local test that uses along + lineSlice + length full circle and the accuracy isn't that great. (correct to 1%-level but not at 0,1%-level) But I read in turf.js some complaints about the accuracy of lineSlice so it might be related to that.

leiflinse-trivector:
I can add that I did some testing with lineSlice, along and length that I wrote both in dart and in javascript to verify the port. So I am quite certain that the error-rate is from turf.js itself rather than the port.

lukas-h:
@leiflinse-trivector I think this is good enough for a merge. 

@jsiedentop any objections? looking forward to the review

leiflinse-trivector:
> @leiflinse-trivector I think this is good enough for a merge.

I agree.

jsiedentop:
Since this project is a port of TurfJs, it is also legetim if we provide the same accuracy. Perhaps it would be good to mention the inaccuracy in the method description? We could also link to the TurfJs issue in the code.

@leiflinse-trivector was it [TurfJs - lineSlice is very imprecise #2023](https://github.com/Turfjs/turf/issues/2023) you were talking about?

lukas-h (approved):

leiflinse-trivector:
> Since this project is a port of TurfJs, it is also legetim if we provide the same accuracy. Perhaps it would be good to mention the inaccuracy in the method description? We could also link to the TurfJs issue in the code.
> 
> @leiflinse-trivector was it [TurfJs - lineSlice is very imprecise #2023](https://github.com/Turfjs/turf/issues/2023) you were talking about?

Yes that is the complaints on turf.js that I was referring to. The tests I did myself have not been published there though.

PR #161, opened 2024-01-26
Change return type of along to Feature<Point>

leiflinse-trivector:
This goes along with preferring to pass around full Features, and it allow to in the future set properties on the returned point without it being a breaking change.


jsiedentop (approved):
LGTM

PR #170, opened 2024-02-17
Polygon clipping package conversion

aardrop:
A dart conversion of [polygon_clipping](https://github.com/mfogel/polygon-clipping/tree/main) with adopted dart native classes such as Point, flutter's Vector2, SplayTreeMap, and SplayTreeSet.

### DIFF
--- a/lib/src/polygon_clipping/bbox.dart
+++ b/lib/src/polygon_clipping/bbox.dart
@@ -1,0 +1,36 @@
+import 'dart:math';
+
+class BoundingBox {
+  Point ll; // Lower left point
+  Point ur; // Upper right point
+
+  BoundingBox(this.ll, this.ur);
+}
+
+bool isInBbox(BoundingBox bbox, Point point) {
+  return (bbox.ll.x <= point.x &&
+      point.x <= bbox.ur.x &&
+      bbox.ll.y <= point.y &&
+      point.y <= bbox.ur.y);
+}
+
+BoundingBox? getBboxOverlap(BoundingBox b1, BoundingBox b2) {
+  // Check if the bboxes overlap at all
+  if (b2.ur.x < b1.ll.x ||
+      b1.ur.x < b2.ll.x ||
+      b2.ur.y < b1.ll.y ||
+      b1.ur.y < b2.ll.y) {
+    return null;
+  }
+
+  // Find the middle two X values
+  final lowerX = b1.ll.x < b2.ll.x ? b2.ll.x : b1.ll.x;
+  final upperX = b1.ur.x < b2.ur.x ? b1.ur.x : b2.ur.x;
+
+  // Find the middle two Y values
+  final lowerY = b1.ll.y < b2.ll.y ? b2.ll.y : b1.ll.y;
+  final upperY = b1.ur.y < b2.ur.y ? b1.ur.y : b2.ur.y;
+
+  // Create a new bounding box with the overlap
+  return BoundingBox(Point(lowerX, lowerY), Point(upperX, upperY));
+}
--- a/lib/src/polygon_clipping/geom_out.dart
+++ b/lib/src/polygon_clipping/geom_out.dart
@@ -1,0 +1,242 @@
+import 'dart:math';
+
+import 'package:turf/src/polygon_clipping/intersection.dart';
+import 'package:turf/src/polygon_clipping/sweep_event.dart';
+import 'package:turf/src/polygon_clipping/segment.dart';
+import 'package:turf/src/polygon_clipping/vector.dart';
+
+class RingOut {
+  List<SweepEvent> events;
+  PolyOut? poly;
+
+  RingOut(this.events) {
+    for (int i = 0, iMax = events.length; i < iMax; i++) {
+      events[i].segment!.ringOut = this;
+    }
+    poly = null;
+  }
+  /* Given the segments from the sweep line pass, compute & return a series
+   * of closed rings from all the segments marked to be part of the result */
+  static List<RingOut> factory(List<Segment> allSegments) {
+    List<RingOut> ringsOut = [];
+
+    for (int i = 0, iMax = allSegments.length; i < iMax; i++) {
+      final Segment segment = allSegments[i];
+      if (!segment.isInResult() || segment.ringOut != null) continue;
+
+      SweepEvent prevEvent;
+      SweepEvent event = segment.leftSE;
+      SweepEvent nextEvent = segment.rightSE;
+      final List<SweepEvent> events = [event];
+
+      final Point startingPoint = event.point;
+      final List<Intersection> intersectionLEs = [];
+
+      while (true) {
+        prevEvent = event;
+        event = nextEvent;
+        events.add(event);
+
+        if (event.point == startingPoint) break;
+
+        while (true) {
+          List<SweepEvent> availableLEs = event.getAvailableLinkedEvents();
+
+          if (availableLEs.isEmpty) {
+            Point firstPt = events[0].point;
+            Point lastPt = events[events.length - 1].point;
+            throw Exception(
+                'Unable to complete output ring starting at [${firstPt.x}, ${firstPt.y}]. Last matching segment found ends at [${lastPt.x}, ${lastPt.y}].');
+          }
+
+          if (availableLEs.length == 1) {
+            nextEvent = availableLEs[0].otherSE!;
+            break;
+          }
+
+          ///Index of the intersection
+          int? indexLE;
+          for (int j = 0, jMax = intersectionLEs.length; j < jMax; j++) {
+            if (intersectionLEs[j].point == event.point) {
+              indexLE = j;
+              break;
+            }
+          }
+
+          if (indexLE != null) {
+            Intersection intersectionLE = intersectionLEs.removeAt(indexLE);
+            List<SweepEvent> ringEvents = events.sublist(intersectionLE.id);
+            ringEvents.insert(0, ringEvents[0].otherSE!);
+            ringsOut.add(RingOut(ringEvents.reversed.toList()));
+            continue;
+          }
+
+          intersectionLEs.add(Intersection(
+            events.length,
+            event.point,
+          ));
+
+          Comparator<SweepEvent> comparator =
+              event.getLeftmostComparator(prevEvent);
+          availableLEs.sort(comparator);
+          nextEvent = availableLEs[0].otherSE!;
+          break;
+        }
+      }
+
+      ringsOut.add(RingOut(events));
+    }
+
+    return ringsOut;
+  }
+
+  bool? _isExteriorRing;
+
+  bool get isExteriorRing {
+    if (_isExteriorRing == null) {
+      RingOut enclosing = enclosingRing();
+      _isExteriorRing = (enclosing != null) ? !enclosing.isExteriorRing : true;
+    }
+    return _isExteriorRing!;
+  }
+
+  //TODO: Convert type to List<Point>?
+  List<List<double>>? getGeom() {
+    Point prevPt = events[0].point;
+    List<Point> points = [prevPt];
+
+    for (int i = 1, iMax = events.length - 1; i < iMax; i++) {
+      Point pt = events[i].point;
+      Point nextPt = events[i + 1].point;
+      if (compareVectorAngles(pt, prevPt, nextPt) == 0) continue;
+      points.add(pt);
+      prevPt = pt;
+    }
+
+    if (points.length == 1) return null;
+
+    Point pt = points[0];
+    Point nextPt = points[1];
+    if (compareVectorAngles(pt, prevPt, nextPt) == 0) points.removeAt(0);
+
+    points.add(points[0]);
+    int step = isExteriorRing ? 1 : -1;
+    int iStart = isExteriorRing ? 0 : points.length - 1;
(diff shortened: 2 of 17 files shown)


aardrop:
This initial draft commit aimed to set a baseline direct line-by-line conversion of the original JS package to the best of my ability. There’s a lot of work to be done and no functionality has been tested yet, here’s the state of this draft and what’s left to be decided and completed. 

Basic work left to be done:

- Debug and test basic functionality
- Is the class approach what we want to work with?
- Are SplayTreeMap and SplayTreeSet working properly?
- How do we want to stage polygon_clipping to expose object files relative to the original project?
- Check nullsafety implementation and classes relative to original work?
- How do we want to implement points relative to original work? Segment implements a point that also tracks attached events, this is currently being handled with an extension of the original math class.
- Do we want to use this BoundingBox class, the one from Turf, or the Rectangle class from dart that has all the prebuilt method that the original project writes and is maintained under the dart repository?
- What class structuring can we do to improve performance and readability?
- Currently all variable maintain original names for the JS package but does naming convention need to be updated?
- With the heavy typing of dart can we leverage the class structure more with things like factory constructors to improve dart specific performance?
- Are there more generic JS inputs that could be improved by heavier typing like List<List<double> to Turf’s polygon or List<Point>
- How do we want to handle the index file?
    
    

High Level Work Left to be Done:

- test and verify functionality from the bottom up
- Verify approach and it’s fit into the turf_dart project
- Write benchmarks

lib/src/polygon_clipping/rounder.dart
```diff
@@ -0,0 +1,83 @@
+import 'dart:collection';
+import 'dart:math';
+
```
lukas-h, line 3: we already have rounding functions in our helpers file.

Please take a look if that works for you: https://github.com/dartclub/turf_dart/blob/f377e468c283bb36d88520016e8e5eb6a975ab1b/lib/src/helpers.dart#L81

Another thing: Please always `Position` class for coordinates (and maybe an extension or extend to add new things)

lib/src/polygon_clipping/segment.dart
```diff
@@ -0,0 +1,573 @@
+// Give segments unique ID's to get consistent sorting of
+// segments and sweep events when all else is identical
+import 'dart:math';
+
```
lukas-h, line 4: generally: strong types for parameters and return types of functions. Will take a look into this file in a future PR

lib/src/polygon_clipping/vector.dart
```diff
@@ -0,0 +1,26 @@
...
+    basePt.x.toDouble(),
+    basePt.y.toDouble(),
+    endPt2.x.toDouble(),
+    endPt2.y.toDouble(),
+  );
+  return res > 0
+      ? -1
+      : res < 0
+          ? 1
+          : 0;
```
lukas-h, line 25: we already have implemented those functions on `Position`s

lib/src/polygon_clipping/vector_extension.dart
```diff
@@ -0,0 +1,73 @@
...
+import 'package:vector_math/vector_math.dart';
+
+extension Vector2Extension on Vector2 {
+  /* Given a vector, return one that is perpendicular */
+  Vector2 get perpendicularVector {
+    return Vector2(-y, x);
+  }
+}
+
+/* Get the intersection of two lines, each defined by a base point and a vector.
```
lukas-h, line 12: we already have most of the vector maths implemented on `Position` type

lukas-h (changes requested):
The main points from my code review:

–> I didn't take a look at every line of code, but I think I already pointed out a few important things

–> we always need strongly typed parameters and return types of functions (if you have dynamic, consider replacing them with GeoJSONObject [for any GeoJSON object] or GeometryObject [for a concrete Geometry like Point, LineString, Polygon, and the Multi** types])

-> Make sure to always use the types from our library, meaning, coordinates should always use the class `Position` (which contains a lot of vector operations already (addition, subtraction, multiplication, etc.)) and `Point` (which is the standard type from the GeoJSON RFC standard to encode geographical points)

-> We already have a sweepline intersections package, please take a look if you can use that instead of implementing all the logic from scratch https://github.com/dartclub/sweepline_intersections/

aardrop:
I'll dive into this today thanks! I'll look at reusing the sweepline though the sweepline from polygon_clipping wasn't very well commented so I still need to understand it's functionality a bit more to understand if they are interchangeable.

lib/src/polygon_clipping/vector.dart
aardrop: I can't seem to find the equivalent of the compareVectorAngles. Do you happen to know which method you are referring to? I see the dot and cross product.

lib/src/polygon_clipping/bbox.dart
```diff
@@ -0,0 +1,36 @@
...
+  final lowerX = b1.ll.x < b2.ll.x ? b2.ll.x : b1.ll.x;
+  final upperX = b1.ur.x < b2.ur.x ? b1.ur.x : b2.ur.x;
+
+  // Find the middle two Y values
+  final lowerY = b1.ll.y < b2.ll.y ? b2.ll.y : b1.ll.y;
+  final upperY = b1.ur.y < b2.ur.y ? b1.ur.y : b2.ur.y;
+
+  // Create a new bounding box with the overlap
+  return BoundingBox(Point(lowerX, lowerY), Point(upperX, upperY));
+}
```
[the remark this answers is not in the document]
aardrop: I see that method, could I add methods and helpers to the existing BBox method to make it more interoperable with the rest of the package's dependency on Positions? 

i.e. add factory constructor from Positons, add getter methods for Position 1 and 2 of the List<num> lat lng points described?

lib/src/polygon_clipping/rounder.dart
aardrop: Is it worth running a benchmark on the two methods or should I just use the existing one in helpers?

lib/src/polygon_clipping/vector_extension.dart
aardrop: I'm looking at the Position; I need help finding some of the methods needed for this package. Would you like me to add them to the position class or continue using the vector package from vector math, which is supported and maintained by the Flutter and Flare teams?  

My gut reaction makes me nervous about treating Position like a point and a vector, but it might make little difference in implementation. I'll follow your lead, so please let me know what works best for turf.

lib/src/polygon_clipping/bbox.dart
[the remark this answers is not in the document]
aardrop: One more question on the bounding box, the specification the BBox class references https://datatracker.ietf.org/doc/html/rfc7946#section-5

It shows the convention of the first point being the lower left and the second point being the upper right but the BBox does not require these conditions but this polygon clipping uses this relationship to adjust the bounds to the points of the polygons. 

Do you want me to add this to the BBox class, add methods/getters to make BBox work for polygon clipping, or make an extended class with this relationship?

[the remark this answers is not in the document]
lukas-h: > I see that method, could I add methods and helpers to the existing BBox method to make it more interoperable with the rest of the package's dependency on Positions?
> 
> i.e. add factory constructor from Positons, add getter methods for Position 1 and 2 of the List lat lng points described?

sounds good – go for it!



> Do you want me to add this to the BBox class, add methods/getters to make BBox work for polygon clipping, or make an extended class with this relationship?

let's stay conformant with the RFC and maybe work with an extension or class that extends BBox

lib/src/polygon_clipping/rounder.dart
lukas-h: If the existing one works for this use case, let's use that one. And if you see a need for perf. optimization, you can go for it and propose a rewrite of the round we have

lib/src/polygon_clipping/vector.dart
lukas-h: ok, `compareVectorAngles` is the one function we don't have, but we have most of the others (esp. basic vector arithmetic like + - )

You can just use:
`Position(...) + Position(...)` – it uses operator overloading in dart

lib/src/polygon_clipping/bbox.dart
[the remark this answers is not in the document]
aardrop: Sounds good to clarify the RFC seems to follow the lower left upper right convention, so we want to enforce that?

lib/src/polygon_clipping/vector_extension.dart
lukas-h: is there a difference in vector and coordinate/position in the package you're working in? Because basically, a Position is a geodetic coordinates, which is a spherical coordinate. It's not a cartesian coordinate though.

you can just do something like (maybe):
```dart
extension PositionExt on Position {
  /* Given a vector, return one that is perpendicular */
 Position get perpendicularVector {
    return Position.named(lng: -lat, lat: lng);
  }
}
```

lukas-h: important to notice too is that altitude (the third dimension) in our Position is optional. So most of the times we have a 2D coordinate

aardrop: In this implementation, I don't think there's a major difference but I followed this package line by line to the best of my ability https://github.com/mfogel/polygon-clipping/tree/main

I'm well aware of the geodetic coordinate system and, but this package comments on it beyond what it says in the read me. The package distinguishes between the two events, though they only use the "x" and "y" components of both. I'm well-versed in vector math but don't see any reason we can't just use Position for both other than the naming convention unless we need or want to use the features maintained by vector_math down the line. 

In short we definitely don't need to have vectors from what I can tell right now.

lib/src/polygon_clipping/bbox.dart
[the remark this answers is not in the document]
lukas-h: yes

aardrop:
> The main points from my code review:
> 
> –> I didn't take a look at every line of code, but I think I already pointed out a few important things
> 
> –> we always need strongly typed parameters and return types of functions (if you have dynamic, consider replacing them with GeoJSONObject [for any GeoJSON object] or GeometryObject [for a concrete Geometry like Point, LineString, Polygon, and the Multi** types])
> 
> -> Make sure to always use the types from our library, meaning, coordinates should always use the class `Position` (which contains a lot of vector operations already (addition, subtraction, multiplication, etc.)) and `Point` (which is the standard type from the GeoJSON RFC standard to encode geographical points)
> 
> -> We already have a sweepline intersections package; please take a look if you can use that instead of implementing all the logic from scratch https://github.com/dartclub/sweepline_intersections/

I looked at the sweepline_intersections package this weekend, and there might be a way to use this, but its logic is different than the one used by polygon_clipping, and as far as I can tell, it wouldn't be an easy swap with the logic already written. I'm happy to switch, but I'd probably need support from someone who understands that package better. 

The [algorithm used](https://github.com/mfogel/polygon-clipping/blob/HEAD/paper.pdf) for the intersections and the optimization of construction between the intersections is deeply intertwined between the relationships of the segmenting and reconstruction of polygon_clipping. Because the "intersection" is more than just a point in polygon clipping, returning both the broken segments and their relationships to each other and the optimization of the algorithm. If someone more familiar with that package wants to look at the relationship between SweetEvents, SweepLines, and Segments in this package, I'm happy to change it, but I don't see a good way to implement it.

lukas-h:
> written. I'm happy to switch, but I'd probably need support from someone who understands that package better.

We can start with what you already have. Maybe @jsiedentop would be a good conversation partner to talk about different sweepline algo implementations.

jsiedentop:
@lukas-h I'm not yet as deep into the subject as you may think.

I would be fine with using different implementations initially. We could create an issue on the topic so that we can take a closer look later. It seems, that we have potential to improve the sweetline_intersections package to fit our needs in this case.

With good test coverage of the polygon_clipping functionality, we shouldn’t be afraid to change the implementation under the hood someday in the future.

lib/src/polygon_clipping/geom_out.dart
```diff
@@ -0,0 +1,242 @@
...
+    List<List<List<List<double>>>> geom = [];
+
+    for (int i = 0, iMax = polys.length; i < iMax; i++) {
+      List<List<List<double>>>? polyGeom = polys[i].getGeom();
+      if (polyGeom == null) continue;
+      geom.add(polyGeom);
+    }
+
+    return geom;
+  }
```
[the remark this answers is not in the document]
aardrop: Agreed, any advice on deliniating the return between multipolygon or polygons? I'm making good progress on testing all the components and working my way up to these here now.

jsiedentop:
@aardrop  I'm planning a new release for this week. Any chance to finish this feature or will we save it for the next release?

aardrop:
> @aardrop  I'm planning a new release for this week. Any chance to finish this feature or will we save it for the next release?

I'd love to finish it this week but I'm just over half way done writing the testing scripts. Unfortunately because it's all interconnected I'm not even certain its functional until I finish these tests. I'd plan on this going in the next one to be safe. If I can pull this together in the next two days I'll let you know!

jsiedentop:
Thanks for the update. Take your time with the testing – there's no rush. I've planned a few smaller topics for the next days, so just wanted to check if we should sync up for the release or if you need more time.

PR #174, opened 2024-02-27
Implement lineOverlap, booleanOverlap, refactoring and bug fixes

jsiedentop:
Implementation completed so far. The pull request is waiting for a new version of turf_equality to be published. I will change the `pubspec.yaml` as soon as the new version is available. In the meantime, @lukas-h you could do a review if you like.


lib/turf.dart
```diff
@@ -15,3 +15,5 @@ export 'src/midpoint.dart';
 export 'src/nearest_point.dart';
 export 'src/polyline.dart';
 export 'src/nearest_point_on_line.dart';
+export 'boolean.dart';
+export 'src/line_overlap.dart';
```
lukas-h, line 19: maybe we can use the other files in lib/ too?

lib/src/booleans/boolean_overlap.dart
```diff
@@ -0,0 +1,179 @@
...
+///
+/// final isOverlapping = booleanOverlap(first, second);
+/// final isNotOverlapping = booleanOverlap(second, third);
+/// ```
+bool booleanOverlap(
+  Feature firstFeature,
+  Feature secondFeature,
+) {
+  var first = getGeom(firstFeature);
+  var second = getGeom(secondFeature);
```
lukas-h, line 62: final?

lukas-h (approved):
final instead of var?

PR #183, opened 2024-04-09
Add: simplify for Feature<LineString>, ported from turf.js

leiflinse-trivector:
This ports over [simplify](https://github.com/Turfjs/turf/blob/master/packages/turf-simplify/lib/simplify.js) from turf.js.

In the current form it only supports `Feature<LineString>` while the original appears to also support Polygons. Would like some advice on that.


test/components/simplify_test.dart
```diff
@@ -0,0 +1,85 @@
...
+  );
+}
+
+List<Position> _roundCoords(List<Position> coords, num precision) {
+  return coords
+      .map((p) => Position(_round(p.lng, precision), _round(p.lat, precision)))
+      .toList();
+}
+
+num _round(num value, num precision) {
```
lukas-h, line 83: we have a `round` function in helpers:
https://github.com/dartclub/turf_dart/blob/a80aeafa333f8ad33096c8f37c2288b03622b392/lib/src/helpers.dart#L81

lukas-h:
hey @leiflinse-trivector, any updates on this? :)

PR #194, opened 2024-10-17
fix wrong conversion from meters to yards and from meters to inches

dimonkomaron:
1. Fixed wrong conversion for yards
2. Use more precise conversion from meters to inches


lukas-h (approved):
looks good

PR #212, opened 2025-04-03
Implement polygonTangent

deanpapas:
These changes aim to implement and test a new feature called polygonTangent, which calculates tangent lines from an external point to a polygon.


lib/src/polygon_tangents.dart
```diff
@@ -0,0 +1,179 @@
...
+/// );
+///
+/// // Create a point
+/// final point = Point(coordinates: Position.of([61, 5]));
+///
+/// // Calculate tangents
+/// final tangents = polygonTangents(point, polygon);
+///
+/// // The FeatureCollection 'tangents' now contains the two tangent points.
+///
```
lukas-h, line 42: close the codeblock with the three ```

lib/src/polygon_tangents.dart
```diff
@@ -0,0 +1,179 @@
...
+
+FeatureCollection polygonTangents(Point point, GeoJSONObject inputPolys) {
```
lukas-h, line 44: can we specify the type to be FeatureCollection<Point>?

lukas-h (changes requested):
minor changes

lib/src/polygon_tangents.dart
deanpapas: While I could update the method to accept FeatureCollections instead of a GeoJSONObject, that would mean iterating over the entire collection, verifying the geometry type for each feature, running the tangent algorithm on each polygon, and returning the result for each feature. Because this function currently scoped to expect a single feature—specifically a Polygon or MultiPolygon—it makes more sense to allow any GeoJSONObject and then enforce the type at runtime:

Example:
`FeatureCollection polygonTangents(Point point, GeoJSONObject inputPolys) {

  if (inputPolys is! Feature<Polygon> && inputPolys is! Feature<MultiPolygon>) {
    throw Exception("Input must be a Polygon or MultiPolygon feature.");
  }`

lukas-h: Hey @deanpapas my original comment was supposed to say `FeatureCollection<Point>` but the `<Point>` generic type was omitted by github's markdown formatting

lukas-h: > @lukas-h  While I could update the method to accept FeatureCollections instead of a GeoJSONObject,

I am quite happy with the input parameter types, I was trying to talk about the return type, which we could specify as `FeatureCollection<Point>` I think

PR #216, opened 2025-04-03
Point on feature final

### LINKED ISSUE #200: pointOnFeature

Newmanjack:
These changes aim to implement and test a new feature called pointOnFeature, which returns a GeoJSON Feature<Point> guaranteed to be on the surface of the input feature. This implementation replicates Turf.js functionality and supports multiple geometry types.


benchmark/point_on_feature_benchmark.dart
```diff
@@ -0,0 +1,65 @@
+import 'package:benchmark/benchmark.dart';
+import 'package:turf/turf.dart';
+
+// Create some test features for benchmarking
+var point = Feature(
```
lukas-h, line 5: use final instead of var if value is only assigned once

lib/src/point_on_feature.dart
```diff
@@ -0,0 +1,214 @@
+import 'dart:math' as math;
+import 'package:geotypes/geotypes.dart'; // We still need the GeoJSON types, as they're used throughout the package
+
+/// Returns a Feature<Point> that represents a point guaranteed to be on the feature.
```
lukas-h, line 4: make sure we are using the proper dart doc comment syntax (wrapping types in [ ] e.g.)

https://dart.dev/effective-dart/documentation

lib/src/point_on_feature.dart
```diff
@@ -0,0 +1,214 @@
...
+      if (geometry.coordinates.isNotEmpty && geometry.coordinates.first.isNotEmpty) {
+        final firstPoly = Polygon(coordinates: geometry.coordinates.first);
+        return pointOnFeature(Feature(
+            geometry: firstPoly, properties: featureInput.properties));
+      }
+    }
+  }
+  
+  // Unsupported input type.
+  return null;
```
lukas-h, line 77: let's make this function non-nullable and throw an exception instead. what do you think?

lib/src/point_on_feature.dart
```diff
@@ -0,0 +1,214 @@
...
+}
+
+/// Calculates the arithmetic centroid of a Polygon's outer ring.
+Point calculateCentroid(Polygon polygon) {
```
lukas-h, line 81: we already have a centroid function – can you check if we can use that instead? This would allow us to have more code sharing

lukas-h: less code duplication

lib/src/point_on_feature.dart
```diff
@@ -0,0 +1,214 @@
...
+  final count = outerRing.length;
+  for (final pos in outerRing) {
+    sumX += pos[0] ?? 0.0;
+    sumY += pos[1] ?? 0.0;
+  }
+  return Point(coordinates: Position(sumX / count, sumY / count));
+}
+
+/// Calculates a representative midpoint on a LineString.
+Feature<Point> _midpointOnLine(LineString line, Map<String, dynamic>? properties) {
```
lukas-h, line 94: there should be a similar function for midpoint that we could use

lib/src/point_on_feature.dart
```diff
@@ -0,0 +1,214 @@
...
+  final midY = (start[1] ?? 0.0) + ((end[1] ?? 0.0) - (start[1] ?? 0.0)) / 2;
+  
+  return Feature<Point>(
+    geometry: Point(coordinates: Position(midX, midY)),
+    properties: properties
+  );
+}
+
+/// Checks if a point is inside a polygon using a ray-casting algorithm.
+bool _pointInPolygon(Position point, Polygon polygon) {
```
lukas-h, line 128: there is also already an extra pip library that we might be able to use

lukas-h (changes requested):
I found a lot of potential to share code with already existing functions of this lib. Can you check that out? :)

lib/src/point_on_feature.dart
aardrop: Agreed, error handling with messaging including "unsupported input type (type of input)"

lib/src/point_on_feature.dart
aardrop: So turf has updated a lot recently, it would be good to compare reference code for the optimal version. Turf JS also breaks out many methods to be isolated so I think we need to decide if we want to do that or anticipate pulling the whole package and trusting the compiled to remove unused code.

lukas-h (approved):
Approved!

lukas-h:
@GuannanMa please just mention this implementation in the Progress.md, then you can squash-merge yourself without me in the loop!

CC @hamishdgx

PR #221, opened 2025-04-17
implement centerOfMass feature

### LINKED ISSUE #198: centerOfMass
https://github.com/mapbox/concaveman
https://github.com/Turfjs/turf/blob/master/packages/turf-convex/index.ts
https://github.com/Turfjs/turf/blob/master/packages/turf-center-of-mass/index.ts

tiinsy:
More test functions could be implemented including file testing (just unit testing at the moment)


lib/src/center_of_mass.dart
```diff
@@ -0,0 +1,61 @@
+//import 'helpers.dart';
+//import 'invariant.dart';
+import 'package:turf/meta.dart';
+import 'centroid.dart';
+
+Feature<Point> centerOfMass(
```
lukas-h, line 6: we need a doc comment to describe what this function does

lib/src/center_of_mass.dart
```diff
@@ -0,0 +1,61 @@
+//import 'helpers.dart';
```
lukas-h, line 1: remove commented out imports

lib/src/center_of_mass.dart
```diff
@@ -0,0 +1,61 @@
...
+//import 'invariant.dart';
+import 'package:turf/meta.dart';
+import 'centroid.dart';
+
+Feature<Point> centerOfMass(
+  GeoJSONObject geoJson,
+  {Map<String, dynamic>? properties = const {}, }) 
+
+  {
+  List<List<double>> coords =[];
```
lukas-h, line 11: `List<Position> coords` -> we have this `Position` type instead of `List<double>` in turf dart for better typisation

test/components/center_of_mass_test.dart
```diff
@@ -0,0 +1,80 @@
+import 'package:turf/center_of_mass.dart';
+import 'package:turf/meta.dart';
+import 'package:test/test.dart';
+import 'package:turf/src/center_of_mass.dart';
+
+void main () {
```
lukas-h, line 6: are those all the tests from Turf JS?

If there are more, incl. file testing, then it would be great to  implement those too

lukas-h (changes requested):

lukas-h:
did you take a look at my old comments? Is this resolved? @FelAmore @hamishdgx ?

Please click 

Screenshot 2026-03-30 at 08 01 12

hamishdgx:
@FelAmore Looks like a test is failing aswell as analyze is failing. Could you look? If its a false-positive let me know

lib/src/center_of_mass.dart
hamishdgx: Looks like this was unresolved. @FelAmore can you check on it

hamishdgx (approved):

PR #228, opened 2025-05-04
[Feature] added intersection feature

NithyaNN3:


lib/src/intersect.dart
```diff
@@ -0,0 +1,74 @@
+import 'package:geotypes/geotypes.dart';
+import 'package:turf/meta.dart';
+import 'package:polyclip-dart/polyclip.dart'; // Polyclip-dart import (to be fixed)
```
hamishdgx, line 3: Also relies on polyclip

PR #229, opened 2025-05-04
[Feature] created union file

NithyaNN3:


lib/src/union.dart
```diff
@@ -0,0 +1,70 @@
+import 'package:geotypes/geotypes.dart';
+import 'package:turf/meta.dart';
+import 'package:polyclip-dart/polyclip.dart'; // TSubject to change as per polyclip-dart changes
```
hamishdgx, line 3: This is reliant on the polyclip dependency, which im guessing is this - https://github.com/dartclub/turf_dart/pull/170/changes#diff-a6e1111b4e7dc44912802132f53654fcfb28bdae0d5789daa1e8e0821ac18bc4
Although it would be incorrect syntax anyway.

PR #237, opened 2025-05-22
Polygonize function 

### LINKED ISSUE #256: polygonize

Newmanjack:
The `polygonize` function converts a collection of LineString features into a collection of Polygon features by detecting closed rings and properly organizing them into polygons with holes. Here's a detailed explanation of how it works:

## Algorithm Overview

1. __Graph Construction__:

   - The algorithm begins by building a planar graph where each line segment becomes an edge, and line intersections become nodes.
   - Edges are stored with direction information and bearings for efficient traversal.
   - Duplicates are avoided through canonical edge keys based on coordinates.

2. __Ring Detection__:

   - Starting from any edge, the algorithm traverses the graph using the "right-hand rule" (always taking the rightmost turn).
   - This ensures consistent orientation as we walk along edges to form closed rings.
   - Each edge is used exactly once, and the process continues until all edges are processed.

3. __Ring Classification__:

   - Detected rings are then classified as either exterior boundaries or holes.
   - This is done by analyzing spatial relationships between rings.
   - Exterior rings are oriented counter-clockwise per RFC 7946 standard.
   - Interior rings (holes) are oriented clockwise.

4. __Polygon Construction__:

   - The classified rings are assembled into proper polygon structures.
   - Each polygon consists of one exterior ring and zero or more interior rings (holes).
   - Coordinates retain their altitude values if present.

## Technical Innovations

- __Bearing-based Traversal__: Uses pre-computed edge bearings to make optimal right-hand rule decisions.
- __Null-Safe Coordinate Handling__: All coordinate access uses null-safety patterns to handle malformed inputs.
- __Efficient Point Clustering__: Detects separate polygons without requiring manual disconnection.
- __RFC 7946 Compliant__: Follows GeoJSON specification for ring orientation.

## Performance Considerations

- Edge graph construction is O(n) where n is the number of line segments.
- Ring traversal is optimized to O(e) where e is the number of edges.
- Memory footprint is minimized by processing rings sequentially.

This implementation successfully handles all test cases including disjoint polygons, polygons with holes, and preserves altitude values where present.


GuannanMa:
I removed the test-specific special-case handling and further simplified parts of the polygonize implementation by reducing duplicated helper-style logic. I focused mainly on the clustering-related part, since that was the area that seemed the most tied to specific layouts rather than the general graph/ring approach. Tests are still passing locally.

hamishdgx:
@GuannanMa 
```
warning - lib/src/polygonize.dart:15:8 - Unused import: 'package:turf/src/invariant.dart'. Try removing the import directive. - unused_import
warning - lib/src/polygonize.dart:16:8 - Unused import: 'package:turf/src/booleans/boolean_clockwise.dart'. Try removing the import directive. - unused_import
```

GuannanMa:
got it

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

GuannanMa:
I cleaned up the remaining polygonize-related analyzer issues and kept the earlier structural changes. Tests are still passing locally.

lib/src/polygonize/point_clustering.dart
```diff
@@ -0,0 +1,168 @@
+import 'package:turf/helpers.dart';
+import 'position_utils.dart';
+
+/// Utility for clustering points into groups based on proximity
+class PointClustering {
```
hamishdgx, line 5: This isn't called anywhere do we have a test that uses this?

lib/src/polygonize/ring_finder.dart
```diff
@@ -0,0 +1,156 @@
...
+      final angleB = (b.bearing - incomingBearing + 360) % 360;
+      return angleA.compareTo(angleB);
+    });
+
+    // Return the edge with the smallest clockwise angle (right-hand rule)
+    return candidates.first.edge;
+  }
+
+  /// Calculate bearing between two positions
+  num _calculateBearing(Position start, Position end) {
```
hamishdgx: This is duplicated here - https://github.com/dartclub/turf_dart/pull/237/changes#diff-1191c27d71dd1e4c99647ef7afedf9f58735e83e76f8a37efbc11aa7bd100c26R112
Both are doing slightly different things. Is that expected?

hamishdgx: Same with _degreesToRadians and _radiansToDegrees

GuannanMa:
I’ll check whether PointClustering is actually used anywhere in the current flow, and I’ll also clean up the duplicated bearing / radians helper logic in ring_finder.dart so it doesn’t diverge from the existing implementation.



Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

lib/src/polygonize/graph.dart
```diff
@@ -0,0 +1,142 @@
...
+    num lng2 = _degreesToRadians(end[0]!);
+    num lat1 = _degreesToRadians(start[1]!);
+    num lat2 = _degreesToRadians(end[1]!);
+    num a = sin(lng2 - lng1) * cos(lat2);
+    num b = cos(lat1) * sin(lat2) - sin(lat1) * cos(lat2) * cos(lng2 - lng1);
+
+    // Convert to azimuth (0-360°, clockwise from north)
+    num bearing = _radiansToDegrees(atan2(a, b));
+    return (bearing % 360 + 360) % 360; // Normalize to 0-360
+  }
```
hamishdgx:
```suggestion
  double _calculateBearing(Position start, Position end) {
    final lng1 = _degreesToRadians(start[0]!);
    final lng2 = _degreesToRadians(end[0]!);
    final lat1 = _degreesToRadians(start[1]!);
    final lat2 = _degreesToRadians(end[1]!);
    final a = sin(lng2 - lng1) * cos(lat2);
    final b = cos(lat1) * sin(lat2) - sin(lat1) * cos(lat2) * cos(lng2 - lng1);

    // Convert to azimuth (0-360°, clockwise from north)
    final bearing = _radiansToDegrees(atan2(a, b));
    return (bearing % 360 + 360) % 360; // Normalize to 0-360
  }
```

lib/src/polygonize/graph.dart
```diff
@@ -0,0 +1,142 @@
...
+  /// Create a canonical edge key
+  String _createEdgeKey(Position from, Position to) {
+    // Create a key based on the actual coordinate values, not the default toString()
+    final fromKey = '${from[0]},${from[1]}';
+    final toKey = '${to[0]},${to[1]}';
+    return fromKey.compareTo(toKey) < 0 ? '$fromKey|$toKey' : '$toKey|$fromKey';
+  }
+
+  /// Convert degrees to radians
+  num _degreesToRadians(num degrees) {
```
hamishdgx: We need to make sure we're not using num where possible as part of https://github.com/orgs/dartclub/projects/9/views/1?pane=issue&itemId=168473212&issue=dartclub%7Cgeotypes%7C10
This can be double

lib/src/polygonize/graph.dart
```diff
@@ -0,0 +1,142 @@
...
+    return degrees * pi / 180;
+  }
+
+  /// Convert radians to degrees
+  num _radiansToDegrees(num radians) {
```
hamishdgx: double

lib/src/polygonize/point_clustering.dart
```diff
@@ -0,0 +1,168 @@
...
+      final distanceSquared = dx * dx + dy * dy;
+      return PointWithDistance(p, distanceSquared);
+    }).toList();
+
+    // Sort by distance
+    pointsWithDistance
+        .sort((a, b) => a.distanceSquared.compareTo(b.distanceSquared));
+
+    // Check if points form two distinct groups by distance
+    num totalDist = 0;
```
hamishdgx: final

lib/src/polygonize/position_utils.dart
```diff
@@ -0,0 +1,70 @@
...
+
+    return Position.of([x, y]);
+  }
+
+  /// Sort nodes in counter-clockwise order around their centroid (for RFC 7946 compliance)
+  static List<Position> sortNodesCounterClockwise(List<Position> nodes) {
+    if (nodes.isEmpty) return [];
+
+    // Calculate the centroid of all nodes
+    num sumX = 0;
```
hamishdgx: These could be final if you don't set them first 
eg.
```
-    num sumX = 0;
-    num sumY = 0;
-    for (final node in nodes) {
-      sumX += node[0] ?? 0;
-      sumY += node[1] ?? 0;
-    }
+    final sumX = nodes.fold<num>(0, (s, node) => s + (node[0] ?? 0));
+    final sumY = nodes.fold<num>(0, (s, node) => s + (node[1] ?? 0));
```

hamishdgx (changes requested):
Could you go through and see what can be set as final?

hamishdgx (approved):

lib/src/polygonize/graph.dart
```diff
@@ -0,0 +1,142 @@
...
+      edgesByVertex[fromKey] = [];
+    }
+
+    // Calculate bearing for the edge
+    final bearing = _calculateBearing(from, to);
+    edgesByVertex[fromKey]!.add(EdgeWithBearing(Edge(from, to), bearing));
+  }
+
+  /// Calculate bearing between two positions
+  double _calculateBearing(Position start, Position end) {
```
lukas-h: reuse https://github.com/dartclub/turf_dart/blob/main/lib/src/bearing.dart

lib/src/polygonize.dart
```diff
@@ -0,0 +1,39 @@
+import 'package:turf/helpers.dart';
```
lukas-h, line 1: mention this in `Progress.md`

lib/src/polygonize/polygonize.dart
```diff
@@ -0,0 +1,111 @@
...
+      return geometry.coordinates;
+    } else if (geometry is MultiPoint) {
+      return geometry.coordinates;
+    } else if (geometry is MultiLineString) {
+      return geometry.coordinates;
+    } else if (geometry is MultiPolygon) {
+      return geometry.coordinates;
+    }
+    throw ArgumentError('Unknown geometry type: ${geometry.type}');
+  }
```
lukas-h: there should already be a getCoords method somewhere in turf_dart - if you find it - please reuse

lukas-h: ... if it has the same behaviour (esp. in how it treats nested structures like polygons, multipolygons, etc.)

lib/src/polygonize/position_utils.dart
```diff
@@ -0,0 +1,66 @@
...
+        source[1] ?? 0.0,
+        source[2] ?? 0.0,
+      ]);
+    } else {
+      return Position.of([
+        source[0] ?? 0.0,
+        source[1] ?? 0.0,
+      ]);
+    }
+  }
```
lukas-h: I think this is not necessary in Dart...

I believe this is a remnant from js where you have to do some kinda checking on if an array contains the correct data

In this library the constructor of Position already enforces everything we need

Imo we can remove the function

lib/src/polygonize/position_utils.dart
```diff
@@ -0,0 +1,66 @@
+import 'package:turf/helpers.dart';
+import 'dart:math';
+
+/// Utility functions for working with Position objects
+class PositionUtils {
```
lukas-h, line 5: I don't think we need to put this into a class namespace. wdyt?

I usually just keep it as global functions, but it is _taste_

lib/src/polygonize/ring_classifier.dart
```diff
@@ -0,0 +1,139 @@
+import 'package:turf/helpers.dart';
+import 'package:turf/src/area.dart';
+import 'package:turf/src/booleans/boolean_clockwise.dart';
+import 'package:turf/src/booleans/boolean_point_in_polygon.dart';
+
+import 'position_utils.dart';
+
+/// Data structure to track ring classification information
+class RingData {
```
lukas-h, line 9: just double check if there is another part of this lib that already creates ring helpers that have similar characteristics

lukas-h (changes requested):

GuannanMa:
will check them this weekend

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

lib/src/polygonize/ring_classifier.dart
lukas-h: @GuannanMa please resolve the rest of the conversations

lukas-h: then I can merge

lukas-h (approved):

lib/src/polygonize.dart
lukas-h: @GuannanMa this one is left

GuannanMa: I checked this one as well — package:turf/helpers.dart is still needed here because this file uses FeatureCollection, Polygon, and GeoJSONObject in the public wrapper signature.

lib/src/polygonize/ring_finder.dart
lukas-h: we can reuse the bearing function that we already have

lib/src/polygonize/ring_finder.dart
```diff
@@ -0,0 +1,156 @@
...
+
+  /// Convert degrees to radians
+  num _degreesToRadians(num degrees) {
+    return degrees * pi / 180;
+  }
+
+  /// Convert radians to degrees
+  num _radiansToDegrees(num radians) {
+    return radians * 180 / pi;
+  }
```
lukas-h: we should also already have those as standard functions of the lib that we can reuse

lukas-h (changes requested):

lukas-h (approved):

lukas-h:
@GuannanMa we can merge

PR #239, opened 2025-05-24
[Feature] clone + dbscan

NithyaNN3:


lib/src/dbscan.dart
```diff
@@ -0,0 +1,136 @@
+import 'package:turf/clone.dart';
+import 'package:turf/distance.dart';
+import 'package:turf/bbox.dart' as turf_bbox;
+import 'package:rbush/rbush.dart';
```
hamishdgx, line 4: FOR RMIT 2026:
I think we will leave this for a few weeks due to complexity. Right @lukas-h ?

lukas-h: we can skip for now. low hanging fruits first

lib/src/clone.dart
```diff
@@ -0,0 +1,106 @@
+
+// Deep clone any GeoJSON object: FeatureCollection, Feature, Geometry, and Properties.
+
+dynamic clone(dynamic geojson) {
```
lukas-h, line 4: tighten types --- e.g. use GeoJSONObject here ...

lib/src/clone.dart
```diff
@@ -0,0 +1,106 @@
...
+    case 'LineString':
+    case 'Polygon':
+    case 'MultiPoint':
+    case 'MultiLineString':
+    case 'MultiPolygon':
+    case 'GeometryCollection':
+      return cloneGeometry(geojson);
+    default:
+      throw ArgumentError('unknown GeoJSON type');
+  }
```
lukas-h, line 24: we might be able to get rid of most of it - we already should have .clone(...) as an overridable function in GeoJSONObject

lib/src/clone.dart
```diff
@@ -0,0 +1,106 @@
...
+}
+
+dynamic cloneProperties(dynamic properties) {
+  if (properties == null) return {};
+
+  final cloned = <String, dynamic>{};
+
+  (properties as Map<String, dynamic>).forEach((key, value) {
+    if (value is Map) {
+      cloned[key] = cloneProperties(value);
```
lukas-h, line 52: we should be able to get rid of all of these clone helpers. this is something js specific that we don't need

lukas-h (changes requested):
- remove unnecessary code
- run tests
- then merge

PR #246, opened 2025-06-04
Add flatten functionality - flattens Multi* geometries to their single counterparts

### LINKED ISSUE #209: flatten

Newmanjack:


hamishdgx:
I've moved the tests into components and added MOST of the tests from turfJS. 
Do we remember why GeometryCollection is not supported????

lib/src/booleans/boolean_contains.dart
```diff
@@ -89,7 +89,9 @@ bool _doBBoxesOverlap(BBox bbox1, BBox bbox2) {
   if (bbox1[0]! > bbox2[0]! ||
       bbox1[2]! < bbox2[2]! ||
       bbox1[1]! > bbox2[1]! ||
-      bbox1[3]! < bbox2[3]!) return false;
+      bbox1[3]! < bbox2[3]!) {
```
lukas-h: I hope this will work - I am a bit sceptical though, because the third parameters (Altitude) are optional in coordinates and bounding boxes. Can you verify this?

lib/src/flatten.dart
```diff
@@ -0,0 +1,68 @@
...
+/// Replicates behavior from: https://turfjs.org/docs/#flatten
+///
+/// Example:
+/// ```dart
+/// var multiLineString = MultiLineString(coordinates: [
+///   [Position(0, 0), Position(1, 1)],
+///   [Position(2, 2), Position(3, 3)]
+/// ]);
+///
+/// var flattened = flatten(multiLineString);
```
lukas-h: use `final` for the examples - in dart you try to mostly use final as opposed to var to make variables immutable / final

Progress.md
```diff
@@ -62,7 +62,7 @@ Dart. This is an on going project and functions are being added once needed. If
 
 - [ ] combine
 - [x] [explode](https://github.com/dartclub/turf_dart/blob/main/lib/src/explode.dart)
-- [ ] flatten
+- [x] flatten
```
lukas-h: link to the impl file

lukas-h (changes requested):

lib/src/booleans/boolean_contains.dart
hamishdgx: I did this based on the analyze output. I'll verify it works now :)

hamishdgx: I think I am going to need a hand with the logic of this. I will book in a time with you.

hamishdgx: https://github.com/dartclub/turf_dart/pull/246/changes/e01ae9d8ab15c24e735a3c8c5e96a870c8ec2af5 - should fix this. From everything I've tested it seems 4- and 6-number BBox values both use the correct longitudes and latitudes like this.

lib/src/booleans/boolean_contains.dart
```diff
@@ -85,11 +85,14 @@ bool _isPolyInPoly(GeoJSONObject geom1, GeoJSONObject geom2) {
   return true;
 }
 
+/// True when [bbox1] contains [bbox2] in longitude and latitude (RFC 7946 order).
 bool _doBBoxesOverlap(BBox bbox1, BBox bbox2) {
```
lukas-h: I think here we also have to take care of 3 dimension coordinates - with altitudes

lib/src/booleans/boolean_helper.dart
```diff
@@ -180,10 +180,10 @@ bool _isLineCrossingThePolygon(LineString line, GeoJSONObject polygon) {
 }
 
 bool _doBBoxesOverlap(BBox bbox1, BBox bbox2) {
-  if (bbox1[0]! > bbox2[0]!) return false;
-  if (bbox1[2]! < bbox2[2]!) return false;
-  if (bbox1[1]! > bbox2[1]!) return false;
-  if (bbox1[3]! < bbox2[3]!) return false;
+  if (bbox1.lng1 > bbox2.lng1) return false;
```
lukas-h, line 184: here, too

lukas-h (changes requested):

lib/src/booleans/boolean_contains.dart
hamishdgx: Addressed. Also made final.

lukas-h (approved):

PR #253, opened 2026-04-20
Attempt at migrating num to double/int

hamishdgx:
- Some tests were required to be modified to manage reading dynamic JSON properties

You can test this locally if you checkout the geotypes branch of the same name. 

Although I will need help managing the geotypes upgrade when that PR is through so that we can remove the dependency_overrides otherwise we get `00:01 +109 -41: Some tests failed`

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -19,3 +19,7 @@ dev_dependencies:
   build_runner: ^2.4.5
   analyzer: ^6.4.0
   benchmark: ^0.3.0
+
+dependency_overrides:
+  geotypes:
+    path: ../geotypes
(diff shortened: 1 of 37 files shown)


hamishdgx:
Goes with https://github.com/dartclub/geotypes/pull/11

pubspec.yaml
```diff
@@ -19,3 +19,7 @@ dev_dependencies:
   build_runner: ^2.4.5
   analyzer: ^6.4.0
   benchmark: ^0.3.0
+
+dependency_overrides:
```
hamishdgx, line 23: obvs not going to be merged like this, just for testing

PR #258, opened 2026-07-10
add missing umbrella exports for `simplify` and `flip`

domyd:
Hi!

I've noticed that `simplify` and `flip` don't have public exports like the others, so when someone imports `turf.dart` it looks like those features are just missing: 

```dart
import 'package:turf/turf.dart' as turf;

turf.simplify(...); // doesn't work, requires extra 'package:turf/simplify.dart' import
turf.flip(...); // doesn't work, requires extra 'package:turf/flip.dart' import
turf.area(...); // works
```

This PR simply adds them to the turf.dart export file :)


lukas-h:
hi @domyd good catch! Let me merge that after the CI passed

PR #260, opened 2026-09-10
Fix Polyline.decode returning wrong coordinates on web

### LINKED ISSUE #259: Polyline.decode returns wrong coordinates on web
# Description

`Polyline.decode` decodes every negative coordinate delta as a large positive
number when compiled to JavaScript.

The zigzag decode in `lib/src/polyline.dart` uses `~(result >> 1)`. Dart maps
`~` and `>>` onto the JavaScript operators, which give an unsigned 32-bit
result, so `~0` is 4294967295 instead of -1 and the sign is lost.

## Reproduce

```dart
import 'package:turf/polyline.dart';

void main() {
  print(Polyline.decode('_p~iF~ps|U_ulLnnqC_mqNvxq`@'));
}
```

| | first coordinate |
| --- | --- |
| expected (and on the VM) | 38.5, -120.2 |
| `dart compile js` / `-p chrome` | 38.5, 42829.47296 |

Nothing throws. `Polyline.encode` is not affected, because its `~` result feeds
into a `& 0x1f` mask.

The existing tests already cover this, but CI only runs `dart test` on the VM,
so it went unnoticed. Six of the 14 polyline tests fail under
`dart test test/components/polyline.dart -p chrome` on `main`.

## Fix

Reverse the zigzag with arithmetic instead, using `~x == -x - 1`:

```dart
(value & 1) != 0 ? -((value + 1) ~/ 2) : value ~/ 2
```

O-Hannonen:
### This PR
- Closes https://github.com/dartclub/turf_dart/issues/259


lukas-h (approved):
lgtm

